PR #224, opened 2023-12-01
Prepare CT for launch

mariabajzek:
Following steps [here](https://docs.google.com/document/d/1ZC4e2ccYACxS0emT1hq25CW_LTH26kVP9YHL8-ZVUWs/edit), except for updating utilities-for-location, since it was not easy to determine which zip codes corresponded to which utilities, and the current representation won't scale well when we have to add a lot more zip codes. I created an Asana [task for that.](https://app.asana.com/0/1205271193902672/1206080952937946/f). For now, I think the user can pick their utility and it will show up regardless of zip code as long as it was in the request.

I also renamed two of the "Eversource" programs to "Energize CT" since Energize CT is the providing authority and they are also applicable for customers of UI.

### DIFF
--- a/data/programs.json
+++ b/data/programs.json
@@ -102,12 +102,12 @@
       "en": "Norwich Public Utilities Cooling & Heating Incentive Pilot Program"
     },
     "url": {
-      "en": "https://norwichpublicutilities.com/residential/chipp/"
+      "en": "https://norwichpublicutilities.com/residential/chipp"
     }
   },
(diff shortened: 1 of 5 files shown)


oyamauchi:
A year later... I'm sorry we never got to this, but we've now launched CT for real at least! (#627)

mariabajzek:
Yay, congrats! Happy to see it!

PR #260, opened 2024-01-11
Add ability for spreadsheet-to-json to automatically add low income thresholds if defined.

datu925:


scripts/incentive-spreadsheet-to-json.ts
```diff
@@ -19,6 +20,11 @@ async function convertToJson(
   file: IncentiveFile,
   strict: boolean,
 ) {
+  if (strict && !(state in LOW_INCOME_THRESHOLDS_BY_AUTHORITY)) {
```
mmezher, line 23: so, for my understanding, in order to use strict mode, low income thresholds must be defined? What about for states that have no low income incentives? (I'm not sure if this is a possible scenario but dont think there's anything state-wide that makes it a requirement)

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -122,6 +154,22 @@ function standardizeDescription(desc: string): string {
...
+function isPlausibleLowIncomeRow(record: Record<string, string>) {
+  if (
+    record.income_restrictions !== undefined &&
+    record.income_restrictions !== ''
+  ) {
+    return true;
+  }
+  if (
+    record.short_description !== undefined &&
+    record.short_description.toLowerCase().includes('income')
```
mmezher, line 166: curious if you've seen this included in all/most of the incentive descriptions with income requirements? Are there any other patterns?

Also worth double checking if we want to be more conservative for low income rows (i.e. if we only want to record them if we're 100% certain they're defined or not)

scripts/incentive-spreadsheet-to-json.ts
datu925: I wanted to gate this behavior and thought the overlap with using strict in terms of use cases was sufficient to just lump them together. I think most people who want to use one will use the other. But there's no reason they can't be decoupled.

I'll add a separate arg to turn the low-income functionality on/off.

scripts/lib/spreadsheet-standardizer.ts
datu925: This popped up a few times in CO – Maria flagged some cases where the short description referenced income eligibility, and Dena or I just hadn't populated the Income Restrictions for some reason. This check will flag those prior to human review.

it's just a warning, so the human can still just ignore them if the thresholds don't exist/shouldn't be defined, etc.

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -94,9 +100,32 @@ export class SpreadsheetStandardizer {
...
+        state,
+      );
+      if (low_income_program) {
+        output.low_income = authorityName;
+      } else if (low_income_program === undefined) {
+        // This is checked up front by the caller rather than
+        // at the row level to avoid spamming error messages.
+      } else {
+        console.log(
+          `Warning: no low-income thresholds found for ${record.id} despite references to income eligiblity in description or income restrictions columns. This either means: 1) it should use the 'default' thresholds, or 2) you need to define thresholds for it, or 3) it's not a low-income row and no action is necessary.`,
```
mmezher, line 115: What is the action for 1)? How would one use the 'default' thresholds, compared to defining a default threshold for it?

Want to make sure the warning message is as actionable and straightforward as possible.

mmezher (approved):
Just a question about the warning message shown; otherwise, it looks good to me! Thanks for separating out the strict and skip low income features.

scripts/lib/spreadsheet-standardizer.ts
datu925: So the answer (and I'm fleshing out the comment as such) is that you just have to fill in 'default' manually for those rows. Which is too bad, since you'll have to do it every time you rerun the script. Not the best experience.

I don't see any better options, at least in the short-term. There really are three scenarios and the script can't know which one to use. I think we can keep this for now and keep it in mind as an area for future improvement – maybe we can find some way to specify in the spreadsheet that a row should use the default thresholds. Or maybe we do away with default thresholds and change the data structure so it's easier to share low-income thresholds across unrelated programs.

PR #262, opened 2024-01-17
Add editorial notes column and better devex for writing outputs.

datu925:
- make state directory if it doesn't exist
- don't write the invalid records if there aren't any, and clear out previous versions if there are
- add an editorial notes column to the mapping, which we added to Colorado to document records that we don't intend to ever show (open to suggestions for a different name). This won't affect existing spreadsheets without this column.

### DIFF
--- a/scripts/incentive-spreadsheet-to-json.ts
+++ b/scripts/incentive-spreadsheet-to-json.ts
@@ -3,6 +3,7 @@ import { parse } from 'csv-parse/sync';
 import fs from 'fs';
 import fetch from 'make-fetch-happen';
 import minimist from 'minimist';
+import path from 'path';
 
 import { LOW_INCOME_THRESHOLDS_BY_AUTHORITY } from '../src/data/low_income_thresholds';
 import { STATE_SCHEMA, StateIncentive } from '../src/data/state_incentives';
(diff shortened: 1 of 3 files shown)


scripts/lib/spreadsheet-mappings.ts
```diff
@@ -30,6 +30,7 @@ export const FIELD_MAPPINGS: { [index: string]: string[] } = {
   other_restrictions: ['Other Restrictions'],
   stacking_details: ['Stacking Details'],
   financing_details: ['Financing Details'],
+  editorial_notes: ['Editorial Notes'],
```
mmezher, line 33: Are we able to add a comment here, or is it clear elsewhere what goes here? Unsure the name is clear enough for folks who aren't familiar, but don't have a better idea for the name.

scripts/incentive-spreadsheet-to-json.ts
```diff
@@ -53,16 +54,26 @@ async function convertToJson(
     }
   });
 
+  const dir = path.dirname(file.filepath);
+  if (!fs.existsSync(dir)) fs.mkdirSync(dir);
```
mmezher, line 58: before, did we document anywhere that a user needed to make a directory? I don't think I saw it in the README.

mmezher (approved):
Actual change LGTM, couple comments about the naming and documentation.

scripts/incentive-spreadsheet-to-json.ts
datu925: It was mentioned – so now I guess that's unnecessary. I've removed those lines. If the user does create it it's not a problem.

scripts/lib/spreadsheet-mappings.ts
datu925: I added a comment. Eventually, this should be added to the data model, which is also where the other fields are documented, but given possible evolutions to how we do that, I don't think it makes sense to do now.

PR #264, opened 2024-01-19
Strengthen cleaning column names

mmezher:
Make cleaning of program names more robust by replacing all punctuation with spaces, then trimming all but one single space. This matches many more column names and prevents us from doing more special casing based on unique ways of formatting.

Tested by running the compiled script on Rhode Island which currently has some formats that were previously unsupported.


datu925:
Can you add a test in `test/scripts`? You can create custom mappings so it should be pretty easy to elicit the behavior. I'm not clear on which column names weren't supported previously and why these changes help, so I think the test would help document that.

datu925 (approved):

PR #265, opened 2024-01-22
Support nested incentive relationships with "any" and "all" logic

mariabajzek:
Implemented per the discussion in [this doc](https://docs.google.com/document/d/11ZCDOdEj0R3HlMqT1k1W17fDJ1GDitYqgMDSmzmMMG0/edit).

**Test Plan**
Ensuring that the existing unit tests still pass (once I changed the JSON of the old relationships to match the new structure). Also added some more complex test relationships to test the new nested structure.


src/lib/incentive-relationship-calculation.ts
```diff
@@ -93,25 +122,50 @@ export function buildRelationshipGraph(data: IncentiveRelationships) {
...
-  if (prerequisiteIds !== undefined) {
-    for (const prerequisiteId of prerequisiteIds) {
-      if (
-        !maps.eligibleIncentives.has(prerequisiteId) &&
-        maps.eligibleIncentives.has(incentiveId)
-      ) {
+  if (typeof prerequisite === 'string') {
+    if (
+      !maps.eligibleIncentives.has(prerequisite) &&
+      maps.eligibleIncentives.has(incentiveId)
```
datu925, line 135: I prefer this as a guard clause at the top since you can deal with this case without knowing anything else. If the base incentive is already ineligible, we don't need to do any more checking.

Same is true with the IsExcluded method below (but I can't comment on that since it's not in the diff :)

src/lib/incentive-relationship-calculation.ts
```diff
@@ -93,25 +122,50 @@ export function buildRelationshipGraph(data: IncentiveRelationships) {
...
+      return false;
+    }
+  } else if ('anyOf' in prerequisite) {
+    for (const child of prerequisite.anyOf) {
+      if (meetsSinglePrerequisite(incentiveId, child, maps)) {
+        return true;
+      }
+    }
+    return false;
+  } else if ('allOf' in prerequisite) {
```
datu925, line 146: nit: now that there are more conditional branches, IMO it's clearer to have the return within each of the branches, even if it's a bit more verbose. Then each branch is self-contained rather than some (and only) some of the branches falling through.

Tangentially related: if the prerequisite is an unexpected case (i.e. not a string, 'anyOf', or 'allOf'), would we rather return true or false? It feels like false is more conservative.

src/lib/incentive-relationship-calculation.ts
```diff
@@ -93,25 +122,50 @@ export function buildRelationshipGraph(data: IncentiveRelationships) {
   return edges;
 }
 
-// Checks the prerequisites for a single incentive.
-export function meetsPrerequisites(
+// Helper to traverse nested prerequisites and check whether they are met for
+// the given incentive.
+function meetsSinglePrerequisite(
```
datu925, line 127: I don't love the name simply because it doesn't feel like there is anything more "single" about this than meetsPrerequisites... both of them are called on a single incentiveID.

Maybe meetsPrerequisiteGraph or satisfiesPrerequisiteGraph? #namingishard

datu925 (changes requested):
Offline we discussed how updating eligibility should account for incentives that are ineligible due to non-relationship factors like owner status, filing status, and income eligibility. But that should happen as a follow-up. Otherwise this LGTM outside of some minor changes.

mariabajzek:
Ooh I think I noticed another bug in makeIneligible: Now that we have OR incentives, all of the dependents don't necessarily become ineligible... I'll fix that when I address the other comments.

src/lib/incentive-relationship-calculation.ts
mariabajzek: Naming is indeed the worst. And actually even though this is talking about a single prereq object, it is really checking the full tree of prerequisites. How about "meetsNestedPrerequisites"?

src/lib/incentive-relationship-calculation.ts
mariabajzek: We have to check that `prerequisite` is a string before checking for it in the map, so for this one it has to stay like this (I can fix it in IsExcluded though).

src/lib/incentive-relationship-calculation.ts
mariabajzek: I didn't think the schema would allow that so I didn't think about it too hard. I think that would be some kind of weird error state so false seems fine with me.

I rewrote this to be a bit more verbose but probably clearer, let me know what you think.

mariabajzek (commented):
I think I'm losing my mind because I already had handled the case I mentioned in my previous comment.

mariabajzek:
We had discussed offline and decided to add checks for membership in `eligibleIncentives` before checking prerequisites for an incentive (so we could avoid doing checks on an incentive that was already ineligible and so that calling functions wouldn't need to know to do those checks). However, in doing so I realized it broke the logic for switching incentives back to being eligible again.

So, I actually moved all of those checks into the calling functions, which I think is actually more correct since checking prerequisites for an incentive is technically a separate action from checking that incentive's current eligibility.

src/lib/incentive-relationship-calculation.ts
oyamauchi: Yeah, I'm fine with relying on the test-time schema validation to prevent unexpected prerequisite shapes from appearing at runtime.

src/lib/incentive-relationship-calculation.ts
```diff
@@ -93,21 +122,42 @@ export function buildRelationshipGraph(data: IncentiveRelationships) {
   return edges;
 }
 
-// Checks the prerequisites for a single incentive.
-export function meetsPrerequisites(
+// Helper to traverse nested prerequisites and check whether they are met for
+// the given incentive.
+function MeetsNestedPrerequisites(
```
oyamauchi, line 127: lowercase m?

oyamauchi (approved):
lgtm!

src/lib/incentive-relationship-calculation.ts
mariabajzek: Good catch, thanks!

PR #266, opened 2024-01-23
Add concept of permanently ineligible incentives

mariabajzek:
If an incentive was marked as ineligible before relationships were checked (e.g. due to income, owner status, etc), the relationships logic should not be allowed to mark it as eligible.

**Test plan**
Added a unit test for a case where an income-ineligible incentive would have been accidentally marked as eligible by the relationships logic previously.

### DIFF
--- a/src/lib/incentive-relationship-calculation.ts
+++ b/src/lib/incentive-relationship-calculation.ts
@@ -4,6 +4,7 @@ import { StateIncentive } from '../data/state_incentives';
 export interface RelationshipMaps {
   eligibleIncentives: Map<string, StateIncentive>;
   ineligibleIncentives: Map<string, StateIncentive>;
+  permanentlyIneligibleIncentives: Set<string>;
   requiresMap: Map<string, Set<string>>;
   requiredByMap: Map<string, Set<string>>;
   supersedesMap: Map<string, Set<string>>;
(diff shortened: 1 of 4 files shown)


src/lib/incentive-relationship-calculation.ts
```diff
@@ -149,6 +150,7 @@ export function makeIneligible(incentiveId: string, maps: RelationshipMaps) {
     for (const dependentId of maps.supersedesMap.get(incentiveId)!) {
       if (
         maps.ineligibleIncentives.has(dependentId) &&
+        !maps.permanentlyIneligibleIncentives.has(dependentId) &&
```
datu925, line 153: Optional: move the check into MakeEligible itself and return early if the incentive is permanently ineligible. That goes back to the question of what's the responsibility of the caller vs the function itself, but in this case I think a function called MakeEligible ought to know about permanent ineligibility. I also think the list of things we expect all the callers to check is getting a little long and I'm worried about someone eventually missing this in the future.

I won't block on it though, so if you have strong opinions that it's better outside, that's fine.

test/lib/incentive-relationships.test.ts
```diff
@@ -165,7 +165,49 @@ test('test incentive relationship and combined max value logic', async t => {
     if (['B', 'E', 'F'].includes(incentive.id)) {
       t.equal(incentive.eligible, true);
     }
-    if (['A', 'C'].includes(incentive.id)) {
+    if (['A', 'C', 'D'].includes(incentive.id)) {
```
datu925, line 168: This is just correcting an unrelated omission, right?

datu925 (approved):

test/lib/incentive-relationships.test.ts
mariabajzek: Yep, I just apparently missed it last time and added it now.

src/lib/incentive-relationship-calculation.ts
mariabajzek: Yeah I totally agree on this one, it's better inside makeEligible. I'm making the switch.

PR #268, opened 2024-01-24
Adding EV incentives for CT based on testing feedback

mariabajzek:
Also fixed the wording of some descriptions to better differentiate between the incentives, as suggested by testing feedback.

**Test Plan**
Updated the low income test fixture since they are now eligible for the new EV incentives as well.

### DIFF
--- a/data/CT/incentives.json
+++ b/data/CT/incentives.json
@@ -256,7 +256,7 @@
       "homeowner"
     ],
     "short_description": {
-      "en": "Rebate off a qualifying heat pump system, dependent on size. Typical heat pump installations will receive approximately $900."
+      "en": "Rebate for a qualifying heat pump system, dependent on size. Typical heat pump installations will receive approximately $900."
     },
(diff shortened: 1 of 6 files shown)


data/CT/incentives.json
```diff
@@ -523,5 +523,145 @@
...
+    }
+  },
+  {
+    "id": "CT-34",
+    "authority_type": "state",
+    "authority": "ct-deep",
+    "type": "pos_rebate",
+    "payment_methods": [
+      "pos_rebate",
+      "rebate"
```
mmezher, line 650: Does this program actually have two ways of paying out the incentive to users (a POS rebate and a later on rebate?)

mmezher (approved):
A few comments and question about payment types

data/CT/incentives.json
mariabajzek: The way I read it was that you could either get the POS rebate or do a mail-in rebate instead.

PR #269, opened 2024-01-24
Add readme for incentive relationships.

mariabajzek:


src/relationships-README.md
```diff
@@ -0,0 +1,36 @@
+# Incentive Relationships Documentation
```
RandomEtc, line 1: Could you link this from the `### Data` section in CONTRIBUTING?

I also wonder if we should start a `/docs` folder to house these? (Thinking: would there be others? e.g. our style guide for descriptions?)

mariabajzek: I added the link to CONTRIBUTING.

I agree it would be good to put this in a better place like a docs folder, so I moved it. For now I don't know what else should go in there but I imagine there will be more things soon.

Do you think we should convert the style guide google doc to markdown and put it in there?

docs/relationships-README.md
```diff
@@ -0,0 +1,36 @@
...
+}
+```
+
+In the above prerequisites,
+
+- The user can only be eligible for incentive_A if they are eligible to claim incentive_B.
+- The user must be eligible to claim **either** incentive_D or incentive_E in order to be eligible for incentive_B.
+- The user must be eligible to claim **both** incentive_D and incentive_E in order to be eligible for incentive_C.
+
+In the above exclusions, if the user is eligible for incentive_X, they cannot be eligible for incentive_Y. When specifying these, you must choose one to be the primary incentive that supersedes the set of incentives listed with it (currently the relationship definition is one-directonal).
```
oyamauchi, line 32: directional

docs/relationships-README.md
```diff
@@ -0,0 +1,36 @@
...
+
+In the above combinations, the user is only eligible to claim $8000 for incentive_A and incentive_B combined, regardless of the individual values or maximums of those incentives.
+
+There is a check in `schemas.test.ts` that looks for cycles in the relationships data and will fail if one is detected. This is to prevent an infinite loop being introdued in the calculation logic.
```
oyamauchi, line 36: introduced

oyamauchi (approved):
Looks good! Just a couple of typos.

I thought maybe it'd be good to mention that the prereqs relationships can be arbitrarily complex, but hopefully that capability won't actually be used :/

docs/relationships-README.md
mariabajzek: Yikes thanks

mariabajzek:
Thanks, I'll add a note that it can be arbitrarily complex if necessary but we don't expect to need more than 2 levels. Hopefully that will be a good hint that if we're doing more than that maybe something has gone wrong.

PR #272, opened 2024-01-24
Split local incentives into city and county types and add support for serving them.

datu925:
City and County are different kinds of local authorities and the logic for matching them is different, so we split them here and add support for serving them in the API (previously they've been filtered out).

Summary of algorithm:
1. When looking up location information, use the zip code to return a county and city for the request, not just a state ID.
2. Require that for any incentive with a county/city authority type, that authority has county and possibly city attributes defined in authorities.json. This is enforced in tests.
3. For county incentives, the request county and authority county need to match. For city incentives, both the county *and* city attributes need to match, because there can be multiple cities with the same name in a state.

I can add documentation in a README about how to do #2, but want to make sure we are okay with this approach first.

### DIFF
--- a/data/authorities.json
+++ b/data/authorities.json
@@ -109,30 +109,45 @@
         "name": "YW Electric Association"
       }
     },
-    "local": {
-      "co-city-and-county-of-denver": {
-        "name": "City and County of Denver"
-      },
+    "city": {
       "co-city-of-boulder": {
-        "name": "City of Boulder"
-      },
-      "co-boulder-county": {
-        "name": "Boulder County"
+        "name": "City of Boulder",
+        "city": "Boulder",
+        "county": "Boulder"
       },
       "co-town-of-avon": {
-        "name": "Town of Avon"
+        "name": "Town of Avon",
+        "city": "Avon",
+        "county": "Eagle"
       },
       "co-town-of-eagle": {
-        "name": "Town of Eagle"
+        "name": "Town of Eagle",
+        "city": "Eagle",
+        "county": "Eagle"
       },
       "co-town-of-erie": {
-        "name": "Town of Erie"
+        "name": "Town of Erie",
+        "city": "Erie",
+        "county": "Weld"
       },
       "co-town-of-vail": {
-        "name": "Town of Vail"
+        "name": "Town of Vail",
+        "city": "Vail",
+        "county": "Eagle"
+      }
+    },
+    "county": {
+      "co-city-and-county-of-denver": {
+        "name": "City and County of Denver",
+        "county": "Denver"
+      },
+      "co-boulder-county": {
+        "name": "Boulder County",
+        "county": "Boulder"
       },
       "co-unincorporated-eagle-county": {
-        "name": "Unincorporated Eagle County"
+        "name": "Unincorporated Eagle County",
+        "county": "Eagle"
       }
     },
     "state": {
@@ -153,7 +168,6 @@
         "name": "CT Department of Energy & Environmental Protection"
       }
     },
-    "local": {},
     "utility": {
       "ct-eversource": {
         "name": "Eversource"
@@ -170,7 +184,6 @@
     }
   },
   "NY": {
-    "local": {},
     "state": {
       "ny-state-of-ny": {
         "name": "State of NY"
@@ -204,7 +217,6 @@
     }
   },
   "RI": {
-    "local": {},
     "state": {
       "ri-oer": {
         "name": "Rhode Island Office of Energy Resources",
@@ -249,7 +261,6 @@
     }
   },
   "VA": {
-    "local": {},
     "state": {},
     "utility": {
       "va-appalachian-power": {
--- a/src/lib/fetch-amis-for-address.ts
+++ b/src/lib/fetch-amis-for-address.ts
@@ -1,6 +1,6 @@
 import { Database } from 'sqlite';
 import { geocoder } from './geocoder';
-import { AMI, IncomeInfo, MFI, ZipInfo } from './income-info';
+import { AMI, GeoInfo, IncomeInfo, MFI } from './income-info';
 
 export default async function fetchAMIsForAddress(
   db: Database,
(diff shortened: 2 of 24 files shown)


src/lib/fetch-amis-for-address.ts
```diff
@@ -17,6 +17,25 @@ export default async function fetchAMIsForAddress(
...
+  const supplemental = await db.get<GeoInfo>(
+    `
+    SELECT 
+        city, 
+        county_name as county
+    FROM zips
+    WHERE zip = ?
+  `,
+    zip,
+  );
```
oyamauchi, line 31: We've run straight into the limits of our zip code data here -- this table has one row per zip, but there are plenty of zips that span city and county borders. (E.g. 12303 -- the data says it's Albany County, NY but a good chunk of it is in Schenectady County too.)

There are at least two threads to pull on:

- Getting hold of a many-to-many mapping between zip codes and cities/counties.
- Figuring out what to do when a user enters a zip that spans cities/counties. Do we return _no_ city/county incentives? Do we return the _union_ of possible city/county incentives? Do we have to introduce some kind of "maybe eligible depending on your exact location" status? (The latter idea has been around for a while, but we've never been forced to actually confront it yet.)

src/lib/fetch-amis-for-address.ts
```diff
@@ -17,6 +17,25 @@ export default async function fetchAMIsForAddress(
...
+  let city: string | undefined;
+  let county: string | undefined;
+  if (supplemental !== undefined) {
```
oyamauchi, line 34: You can just do `if (supplemental)`, or this:

```
const city = supplemental?.city;
const county = supplemental?.county;
```

oyamauchi (changes requested):
I think the zip--county mapping issue is a showstopper, since it's very likely to result in false positive incentive eligibility.

datu925:
This approach was suggested by @RandomEtc in [this doc](https://docs.google.com/document/d/17Dp-eGMS8ttbJBzXVYJmhzk-AolnAkMM_VqVAvu9Y5Y/edit?disco=AAABEgQzlFs) as "a place to start". I can see arguments both ways.

From a strict correctness point, you're definitely right that this will result in false positives. From the same line of thinking though, there are about 100-150 zip codes that cross *state* lines ([link](https://gis.stackexchange.com/questions/53918/determining-which-us-zipcodes-map-to-more-than-one-state-or-more-than-one-city/223445#223445)). In these cases, we're not just serving false positives, we're serving entirely the wrong state. Granted, this is a fairly small set of zip codes so maybe we hand-waved this away as an edge case. As a percentage, many more zip codes cross state/county lines so I think we should invest in better data sources to try to address this. I can spend some time scouting around today.

As an aside, we should document where the geo data comes from in [CONTRIBUTING.md](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/main/CONTRIBUTING.md) and what its assumptions are, since I wasn't aware the zip data was quite so limited (does this problem also apply to census tracts?).

I'm also curious to hear Tom's perspective. I can imagine proceeding despite the limitations – since we can return the name of the city/country in the response, the frontend could show checkboxes for "Show local incentives from X city, Y county", and the user could remove those incentives if they're not accurate. That kind of filter would eventually expand to more of a dropdown if we eventually get better data that gives us multiple cities/counties for a single zip.

Finally, I think having a more principled answer is important for the real CO launch, but I wonder if we want to proceed simply to have a way to see the local CO incentives while beta testing. There are tests that ensure we don't allow city/county incentives for launch states, only beta ones. If we need to get this right even for beta testing, then we should postpone the calculator testing party on Tuesday until we can figure out our path forward here.

src/lib/fetch-amis-for-address.ts
datu925: Thanks, using the `?` syntax, I just inlined the city/county assignments when creating the location object below.

data/authorities.json
```diff
@@ -109,30 +109,45 @@
...
+        "name": "City and County of Denver",
+        "county": "Denver"
+      },
+      "co-boulder-county": {
+        "name": "Boulder County",
+        "county": "Boulder"
       },
       "co-unincorporated-eagle-county": {
-        "name": "Unincorporated Eagle County"
+        "name": "Unincorporated Eagle County",
```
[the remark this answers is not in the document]
datu925: The language on the website is "Unincorporated Eagle County has specified bonuses and additional rebates for it’s [sic] residents and businesses that can be used in addition to Energy Smart rebates through Walking Mountains Science Center." That reflects at least some degree of agency/authority to me. From the Wikipedia article on unincorporated areas: "Most states have granted some form of home rule, so that county commissions (or boards or councils) have the same powers in these areas as city councils or town councils have in their respective incorporated areas."

https://www.walkingmountains.org/sustainability-hub/unincorporated-eagle-county-energy-efficiency-program/

So, while I agree that your proposed distinction could be helpful at some point, I think this is closer to a normal town/city authority than it sounds. This is one of a set of partnerships between Walking Mountains Science Center and towns/areas nearby (other examples include the towns of Vail and Aspen). The reality is that this is more of a "co-authority" partnership, but that would also require evolving the data model in a way that's probably not needed 99% of the time.

src/lib/fetch-amis-for-address.ts
datu925: I think we want better treatment of this to block state calculator launches, but not beta testing. It's important for us to be able to see local incentives sooner rather than later (on Tuesday as currently scheduled). This is the direction I got from Tom when chatting with him on Slack (he's too busy for a more substantive review).

I've added a number of comments indicating this is a limited solution, and I filed https://app.asana.com/0/1204738794846444/1206454407609847 to track the work for getting to a better long-term solution, but can we proceed with this as a stopgap for now?

oyamauchi (approved):
Discussed offline: we can go ahead with this for testing, but not launch. And we'll work on finding the correct many:many zip/county/city data.

PR #274, opened 2024-01-25
Initial progress toward a collected data schema.

datu925:
This is a proof-of-concept PR. It's more of a sketch than something intended to be reviewed and merged. 

The concept is based on this doc: https://docs.google.com/document/d/1cZMhSaSS4Ckus4KDga3oiTol4iLcZRcc15GguRJCcQ4/edit#heading=h.fa4kckfyfzzf.

The PR has:
- schema updates to show how we could easily create a collective schema that's an augmented version of the existing schema (in reality, I think it would be even more understandable to define the collected schema first and then define the API schema as a subset of it)
- a `convert-formats.ts` script that allows converting in between CSV and JSON versions of that schema. In the longer-term, we can export to Google sheets with some of the validations we currently have in place
- 3 files to show that this works. They're probably not that interesting to look at. `schema-friendly-data.csv` is a version of the CO data that was already converted to fit the schema, so we don't have to worry about changing names of columns/fields. `schema-friendly-data.json` is that CSV after it was converted to JSON. `back-to-csv.csv` is the roundtrip back to CSV. I've checked that they are equivalent (not byte-equivalent – something is off with the newlines or maybe quoting, but they are value-equivalent).


datu925:
This proof-of-concept has been broken up into a bunch of actual PRs, so closing now.

PR #277, opened 2024-01-30
Add supported columns, add alphabetization of authorities.

mmezher:
Added two more supported columns from incentive-a-thon spreadsheet, and added alphabetization of authorities' file. Typical json sorting wasn't successful as the key's name was the value to sort by, which wouldn't parse when defining the sort function.

That being said, a generalized alphabetization solution for both programs and authorities is also being experimented with, but this simpler solution is viable due to incentive calculation being sorted by state locations of incentives.

Programs still need to be sorted, and README details this.

### DIFF
--- a/scripts/lib/authority-and-program-updater.ts
+++ b/scripts/lib/authority-and-program-updater.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import { Project } from 'ts-morph';
+
 const project = new Project({
   tsConfigFilePath: 'tsconfig.json',
 });
--- a/scripts/lib/authority-and-program-updater.ts
+++ b/scripts/lib/authority-and-program-updater.ts
@@ -17,60 +17,6 @@ const wordSeparators =
 const capital_plus_lower = /[A-ZÀ-Ý\u00C0-\u00D6\u00D9-\u00DD][a-zà-ÿ]/g;
 const capitals = /[A-ZÀ-Ý\u00C0-\u00D6\u00D9-\u00DD]+/g;
 
-const stateAbbreviationArray: readonly string[] = [
-  'AL',
-  'AK',
-  'AZ',
-  'AR',
-  'CA',
-  'CO',
-  'CT',
-  'DE',
-  'DC',
-  'FL',
-  'GA',
-  'HI',
-  'ID',
-  'IL',
-  'IN',
-  'IA',
-  'KS',
-  'KY',
-  'LA',
-  'ME',
-  'MD',
-  'MA',
-  'MI',
-  'MN',
-  'MS',
-  'MO',
-  'MT',
-  'NE',
-  'NV',
-  'NH',
-  'NJ',
-  'NM',
-  'NY',
-  'NC',
-  'ND',
-  'OH',
-  'OK',
-  'OR',
-  'PA',
-  'RI',
-  'SC',
-  'SD',
-  'TN',
-  'TX',
-  'UT',
-  'VT',
-  'VA',
-  'WA',
-  'WV',
-  'WI',
-  'WY',
-];
-
 function kebabCase(str: string) {
   // replace word starts with space + lower case equivalent for later parsing
   // 1) treat cap + lower as start of new word
@@ -128,8 +74,17 @@ export function createProgramName(
   );
 }
 
+export function sortJsonAlphabeticallyByStateKey(json: StateToAuthorityMap) {
+  const ordered = Object.keys(json).sort().reduce((obj, key) => {
+    obj[key] = json[key];
+    return obj;
+  },  {} as StateToAuthorityMap);
+  return ordered;
+  }
+
 type AuthorityKey = string;
 type ProgramKey = string;
+type StateKey = string;
 type Authority = {
   name: string;
   authority_type: string;
@@ -142,6 +97,10 @@ type ProgramMap = {
 
 type AuthorityMap = {
   [index: AuthorityKey]: Authority;
+};
+
+type StateToAuthorityMap = {
+  [index: StateKey]: AuthorityMap;
 };
 
 export class AuthorityAndProgramUpdater {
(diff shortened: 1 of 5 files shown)


mmezher:
@datu925 , any suggestions/preferences on how to test? This bit of the script currently has a file dependency that is hard to emulate. One option would be to modify the code to take in the auth file name as a parameter (or write file name as param for generalized alphabetical sorting). 

I've already manually tested, but likely want to be able to long-term test the general code in the future anyways.

scripts/lib/authority-and-program-updater.ts
```diff
@@ -157,11 +211,21 @@ export class AuthorityAndProgramUpdater {
...
-    fs.writeFileSync(
-      AUTHORITIES_JSON_FILE,
-      JSON.stringify(json, null, 2),
-      'utf-8',
-    );
+
+    // Reset file first.
+    fs.writeFileSync(AUTHORITIES_JSON_FILE, '', 'utf-8');
+    // Go through list of state abbreviations and then write them, if entries exist.
+    for (const entry_state of stateAbbreviationArray) {
```
datu925, line 218: So I assume this has to do with the weirdness that I didn't understand when we were chatting, but are we sure something like https://stackoverflow.com/questions/5467129/sort-javascript-object-by-key doesn't work?

It seems like we should be able to get the keys of the json, sort them, then iterate over them populating a new object with the key->value pairs. And then write overwrite the original file with that object.

If that doesn't work for some reason and we have to iterate over the abbreviation list, I'd still recommend collecting the stringified states into a single object and then doing a single write. That will ease the testing story (I'll respond to your big-picture testing comment elsewhere).

datu925:
> @datu925 , any suggestions/preferences on how to test? This bit of the script currently has a file dependency that is hard to emulate. One option would be to modify the code to take in the auth file name as a parameter (or write file name as param for generalized alphabetical sorting).
> 
> I've already manually tested, but likely want to be able to long-term test the general code in the future anyways.

I did the expedient thing but I generally think it's better to have heavy-lifting methods operate on data rather than filenames for this reason.

Maybe we have separate the method into an inner and outer method. The outer one reads/writes the file, and the inner one (which is the one that's tested) takes in JSON (or an TypeScript authorities object) and returns JSON (or a TypeScript authorities object). We could also do that as part of the constructor or at least pass in and store the filename in the constructor.

The repeated-appending complicates that separation, but I think we should just do a single write anyway – see my other comment in the file.

datu925 (changes requested):
Couple questions and thoughts about the testing.

scripts/lib/spreadsheet-mappings.ts
```diff
@@ -33,6 +33,8 @@ export const FIELD_MAPPINGS: { [index: string]: string[] } = {
   // This contains notes about why we might not serve a record in the API.
   // It may not appear in all spreadsheets.
   editorial_notes: ['Editorial Notes'],
+  bonus_details: ['Bonus Details'],
```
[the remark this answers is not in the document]
mmezher: Good catch. Agree, dupe.

scripts/lib/authority-and-program-updater.ts
mmezher: Realized https://stackoverflow.com/questions/5467129/sort-javascript-object-by-key will work if I `type` everything; so may not work for programs as-is, but assuming we can either create a separate sort for program's JSON or modify type to take either. Thanks! Will add tests now.

mmezher:
I've added some types in order to enable sorting the objects (yay), but it was more manual work than anticipated. The schema used in the authorities' types file _doesn't_ match exactly what we use/want here, which is why there are new "types" being defined. LMK what you think.

scripts/lib/authority-and-program-updater.ts
```diff
@@ -90,6 +102,18 @@ type AuthorityMap = {
...
+type JsonAuthorities = {
+  [index: AuthorityKey]: { name: string };
+};
+type AuthorityType = string;
+type AuthorityTypeMap = {
+  [index: AuthorityType]: JsonAuthorities;
+};
+export type StateToAuthorityTypeMap = {
+  [index: StateKey]: AuthorityTypeMap;
+};
```
datu925, line 115: Does the `AuthoritiesByState` type in `src/data/authorities.ts` work here instead?

If not, I'm guessing the `AuthoritiesByType` type I made won't help either, but that's also there.

Maybe there's some gotcha but those types are basically these types, so it would be nice to avoid duplicate definitions if possible.

scripts/lib/authority-and-program-updater.ts
```diff
@@ -1,6 +1,5 @@
 import fs from 'fs';
 import { Project } from 'ts-morph';
-
 const project = new Project({
```
datu925, line 3: Doesn't really matter, but is this a prettier change? Otherwise I'd say better to have a line in between imports and everything else.

datu925 (approved):
Approved, but see a couple of my comments first before merging.

scripts/lib/authority-and-program-updater.ts
mmezher: I'm guessing it was an accidental change - prettier doesn't seem to care either way. Adding back.

scripts/lib/authority-and-program-updater.ts
mmezher: Yeah, I experimented extensively with both types, trying to make them fit here. The major issues I've seen are due to us going through the object entries in a fairly unstructured way: e.g. json[stateUpper][authority.authority_type.toLowerCase()] doesn't have any explicit key value for what that authority type should be, so there are many errors thrown unless we do a switch case for each type of possible authority type there (and then, do the same kind of enumeration for API_AUTHORITY_SCHEMA properties as we're writing and creating). 

There's also the issue of possible "undefined" properties when doing it this way for both `AuthoritiesByState` and `AuthoritiesByType`. 

While I do agree reusing the actual types would be extremely beneficial to maintainability, I'd rather do it in a separate change, since it might change the way things are written/defined here altogether. Does that sound reasonable?

datu925 (approved):

PR #282, opened 2024-02-06
Improve spreadsheet-to-json value standardization.

datu925:
This does a few things better than the current approach:

1. It switches VALUE_MAPPINGS to behave more like FIELD_MAPPINGS in several respects. It's now passed in instead of being a global and it's closer to the same format (canonical -> list of aliases).
2. Uses that easier-to-work-with format to rename values as part of standardizing field names. I'll rename that method in a follow-up so as to not touch even more files. As a result, all of the gross enum stuff we had to do can be removed. I think the bracket syntax is temporary; I have a follow-up that will make this all even cleaner.
3. Renames a bunch of columns in `spreadsheet-mappings.ts`. These were done to be closer to the JSON schema (authority_level -> authority_type) or to use a nested syntax like `amount.type` which will make from CSV format to JSON a breeze.

Testing: added unit tests, plus I ran CO through this and I get the exact same values as what's currently in the JSON.


scripts/lib/spreadsheet-mappings.ts
```diff
@@ -36,61 +41,93 @@ export const FIELD_MAPPINGS: { [index: string]: string[] } = {
...
+    heat_pump_water_heater: ['Heat Pump Water Heater (HPWH)'],
+    new_electric_vehicle: ['New Electric Vehicle'],
+    used_electric_vehicle: ['Used Electric Vehicle'],
+    electric_vehicle_charger: ['Electric Vehicle Charger'],
+    rooftop_solar_installation: ['Rooftop Solar'],
+    battery_storage_installation: ['Battery Storage'],
+    heat_pump_clothes_dryer: ['Heat Pump Dryers / Clothes Dryer'],
+    electric_stove: ['Induction Cooktop', 'Electric Stove'],
+    weatherization: ['Weatherization (insulation and air sealing)'],
+    electric_panel: ['Electric panel', 'Electric wiring'],
```
mmezher, line 95: capitalization doesn't matter for these in the end, right? Observing that some columns are all words capitalized, some have only the first letter in the column, some none at all.

If it doesn't matter, should we be consistent throughout - maybe only lowercase?

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -1,53 +1,92 @@
...
-        output[key] = input[key];
-        continue;
+      if (this.fieldMap[cleaned] === undefined && this.strict) {
+        throw new Error(`Invalid column found: ${cleaned}; original: ${key}`);
+      }
+
+      const newCol = this.fieldMap[cleaned] || key;
+      let val = this.convertToCanonical(input[key], newCol);
+
+      // Special case cleaning fields – eventually this can be
```
mmezher, line 78: maybe say more here about how this could improve?

It almost feels like a TODO - is there a format to use for that here?

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -61,38 +100,30 @@ export class SpreadsheetStandardizer {
...
+      item: record.item,
       program: createProgramName(
         state,
         record.authority_name,
         record.program_title,
       ),
-      amount: createAmount(record),
-      owner_status: findOwnerStatus(record.owner_status),
+      amount: this.createAmount(record),
+      owner_status: record.owner_status.split(',').map(x => x.trim()),
```
mmezher, line 114: What does this end up with in the "Both" case? we split on the "homeowner, renter" and then trip blankspaces -- is the end result an array with both? Doesn't this need to transform again into a string? 

Sorry if this is a silly typescript syntax misreading.

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -126,6 +157,35 @@ export class SpreadsheetStandardizer {
...
+    ) {
+      amount.representative = +cleanDollars(input['amount.representative']);
+    }
+    if (
+      input['amount.minimum'] !== undefined &&
+      input['amount.minimum'] !== ''
+    ) {
+      amount.minimum = +cleanDollars(input['amount.minimum']);
+    }
+    if (
```
mmezher, line 178: Seeing this reused a lot - anyway to make into own function able to pass in the input type?

mmezher (commented):
Few questions here and there - generally looks good!

scripts/lib/spreadsheet-standardizer.ts
datu925: I'm going to delete this method in a follow-up anyway once we have a collected data schema.

scripts/lib/spreadsheet-standardizer.ts
datu925: I'm realizing what I really meant by "more principled" is that it's feels a bit awkward as part of that method, so I extracted it to a separate method.

scripts/lib/spreadsheet-mappings.ts
datu925: My current thinking is that the first alias at least is going to be special – when the back-to-csv story is done, that first alias will be treated as the "canonical alias" – that is, the user-friendly string that we allow folks to use when filling out the spreadsheets so they can work with English phrases instead of the machine-friendly stuff.

We could separate those concepts (so that there is a separate "canonical alias" data structure and then these mappings are for the non-canonical aliases), but I think that's just going to result in more mappings to keep track of and I thought it was worse than just using the first-is-canonical convention.

Given that, the capitalization is going to matter. I am game for whatever uppercasing/lowercasing you think is better UX for the spreadsheets. But there's also an argument for keeping it since I assume Preethi chose the current capitalization scheme intentionally.

mmezher: Hm, I see. In that case, I am in support of making the captialization regular (likely every word having a capital first letter?). It doesn't make sense to me that electric panel be "Electric panel" but electric stove be "Electric Stove".

WRT the first alias - I have some concerns that the current groupings (e.g. of "electric stove" and "induction cooktop") will lose valuable data along the way if we're using this list for back-to-csv. Maybe it doesn't matter that much because the description itself will likely include what specifically is being referenced?
(This second comment is not as important for this change)

scripts/lib/spreadsheet-standardizer.ts
mmezher: Ack.

scripts/lib/spreadsheet-standardizer.ts
mmezher: <insert thumbs up here>

mmezher (approved):
More questions about the aliasing - OK to leave any changes there for another change (I can author)

scripts/lib/spreadsheet-mappings.ts
datu925: I'll submit this as is, but do the capitalization as an immediate follow-up since it's pretty simple.

Re: your second comment, I agree with you. This is a fundamental problem with our current heat pump setup – we've got 4 kinds of air source heat pump options in spreadsheets and one in JSON. You can collapse 4 to 1 but you can't expand back to 4 once you've lost that detail.

I didn't want to address that here, but ultimately the alias list should be used only for typos/variations of the exact same technology, whereas this is using it also to collapse things that we don't currently have the vocabulary for in the backend. I'll plan to add those as the immediate follow-up as well.

PR #284, opened 2024-02-07
Add schema for collected data and base our state incentives schema off that.

datu925:
Main thing: adds schemas for collected data and derived data, and then defines downstream schemas like the main StateIncentive in relation to them. We don't exploit that in this PR, but it opens up the door for the features we actually care about, like easier spreadsheet-to-json conversion and tests that assess whether the spreadsheet data itself meets our standards/has problems.

Side thing: I noticed that the amount schema is defined once for state incentives and once for IRA incentives. i think it's best to consolidate until we have a reason to split them.


src/data/state_incentives.ts
```diff
@@ -18,77 +21,142 @@ export type LowIncomeAuthority =
...
+// fields (for example, program_start_raw and program_start). Generally, if
+// what's in the spreadsheet requires more than just fitting to an enum or
+// converting from string to number with minor cleanup, it should have a
+// separate collected and derived version.
+export type DerivedFields = {
   agi_max_limit?: number;
   agi_min_limit?: number;
-  authority_type: AuthorityType;
   authority: string;
   type: PaymentMethod; // Deprecated; we are switching to use payment_methods instead
```
oyamauchi, line 110: This reminds me: we can actually remove this field now (from v1); the frontend is using `payment_methods`. Fine to do that in a separate PR though.

src/data/shared_schemas.ts
```diff
@@ -0,0 +1,16 @@
+import { JSONSchemaType } from 'ajv';
+import { Amount, AmountType, AmountUnit } from './types/amount';
+
+export const AMOUNT_SCHEMA: JSONSchemaType<Amount> = {
```
oyamauchi, line 4: I think this can go in `amount.ts`; there are both schemas and type definitions in there.

oyamauchi (approved):

PR #286, opened 2024-02-12
Integrate ENERGY STAR® utility dataset

oyamauchi:
## Description

This pulls in the mapping from zip codes to utilities in the dataset here:
https://www.energystar.gov/buildings/owners_and_managers/existing_buildings/use_portfolio_manager/find_utilities_provide_data_benchmarking

The centerpiece here is the script that converts that Excel file into
two CSV files that are then imported into the API server's SQLite DB
for runtime querying. It takes the human-friendly utility names from
the sheet, does a bit of cleaning, then turns them into utility IDs in
the same style as we have in the code already
(e.g. `ri-rhode-island-energy`).

The script is not run at build time. It's meant to be run by hand, and
the resulting CSVs checked in, like everything else in `scripts/data`.

For now, the behavior is that the result set is the _intersection_ of
the dataset's result and our own `authorities.json`. This is a
conservative approach, avoiding returning any utility we aren't
already returning today. This may not be the long-term final state; I
think eventually we may not want utilities in `authorities.json` at
all, and instead consider this dataset as canonical.

The user-visible names are still from `authorities.json`, not from the
dataset.

This does mean that we will _stop_ returning some "utilities" that
we're returning today, albeit only in beta states. These are
consortiums of small utilities (e.g. VPPSA in VT, Tri-State G&T in CO)
that offer incentives to customers of their member utilities.
Resolving this problem will entail separating the concept of "who
offers an incentive" from "which utility's customers are eligible".

While going through the dataset's utilities for the states we have
incentive data for, I made some changes to our names/IDs, to better
reflect the way the utilities are branded to customers. The changes in
`data` are all straight find/replace.

## Test Plan

`yarn test`. This includes a new test that ensures all of our IDs in
authorities.json exist in the dataset (except for the exceptions, as
noted above).

Spot-checking with ZIP codes in Rhode Island (where we know the
situation pretty well) and other beta states, especially focusing on
zips around towns where there are municipal utilities.


scripts/generate-utility-data.ts
```diff
@@ -0,0 +1,253 @@
...
+  cleaned = cleaned.replace(/-? *\([A-Z]{2}\)$/, '').trim();
+  // Remove square-bracketed parts, which are parent companies not used as the
+  // customer-facing brand
+  cleaned = cleaned.replace(/\[.*\]/, '').trim();
+  // Remove "Inc" or "Co" suffix
+  cleaned = cleaned.replace(/,? *(I(nc)?|Co)\.?$/, '').trim();
+  // Remove "(for __ reporting)" parenthetical
+  cleaned = cleaned.replace(/\(.*reporting.*\)/i, '').trim();
+  // Spell out "Pwr", "Assn", "Elec", and "Coop"
+  // The dataset is inconsistent about this
```
mmezher, line 126: Without knowing how consistent the dataset is - I notice that some of these replacements require capitalization to match. Should we continue to assume all entries will have capitalization of names? (e.g. below, Pwr will only be cleaned to "Power" if matching exactly).

scripts/generate-utility-data.ts
```diff
@@ -0,0 +1,253 @@
...
+  cleaned = cleaned.replaceAll(/\bPwr\b/g, 'Power');
+  cleaned = cleaned.replaceAll(/\bAssn\b/g, 'Association');
+  cleaned = cleaned.replaceAll(/\bElec\b/g, 'Electric');
+  cleaned = cleaned.replaceAll(/\bCo-?op\b/g, 'Cooperative');
+
+  // Generate the ID: replace & with "and", remove all parentheticals,
+  // spaces into hyphens
+  const id =
```
datu925, line 134: There is an existing createAuthorityName function in `scripts/lib/authority-and-program-updater.ts`. I don't care which we use, but it's important that we have a consistent way to go from authority name to authority ID; otherwise the spreadsheet-to-json flow is going to create invalid IDs.

datu925 (commented):
How do you think this change affects what data collectors do/conversion from spreadsheet to json? A few thoughts:

1. If we're still doing collection in spreadsheets, we want to be able to constrain collectors' choices to the list of valid/recognized utilities. That's probably just dumping the list to one of the templates and having some data validation.
2. Similar in the spreadsheet-to-json flow, we'll want to consider any row with an invalid utility as invalid. I can do that as a follow-up.
3. At some point, we'll want to change the spreadsheets to reflect the name changes you've made here, unless you already did that. It's not urgent since our spreadsheets/JSON are already out of sync, but I will probably add some tests soon that will fail on these kinds of changes.

scripts/generate-utility-data.ts
oyamauchi: Ah, OK. I'll use that but I'll add the `&` -> `and` replacement to it, since that comes up all the time in utility names and I think it makes the resulting ID clearer

scripts/generate-utility-data.ts
oyamauchi: As far as I can tell, the dataset is perfectly consistent about these abbreviations being capitalized. I can make these case-insensitive, but the replacement would still be titlecase. On the whole I think that's better than leaving an abbreviation in, because a human reviewer is more likely to notice something amiss?

oyamauchi (commented):
My rough vision is that when starting collection for a new state, whoever is setting up the spreadsheet would pull all the available utility IDs for the state from this sqlite DB, and set up the sheet's data validation based on those. I think it's even acceptable for the human-facing value to be `co-xcel-energy` rather than `Xcel Energy` -- the former is perfectly human-readable, and it would take away one conversion step.

src/lib/utilities-for-location.ts
```diff
@@ -10,46 +18,31 @@ import { GeoInfo } from './income-info';
...
-        ids = ['ri-rhode-island-energy'];
-    }
-  } else {
-    // Fall back to returning all the state's utilities, regardless of location.
-    return stateUtilities;
-  }
+  // Put the predominant utility first
+  const rows = await db.all<ZipToUtility[]>(
+    'SELECT * FROM zip_to_utility WHERE zip = ? ORDER BY predominant DESC',
+    location.zip,
```
datu925, line 38: If there are no rows found, I think we should return all of the state's utilities in authorities.json as fallback – right now this will just return empty.

Definitely not part of this PR, but what happens if the dataset is wrong and the user is not offered a list that includes their utility? Do they have any way of letting us know/requesting a full state list instead?

It might also be good to do a quick coverage test to make sure this has good coverage across our zips. It looks good overall, but I think that's something to ensure before we go ahead.

datu925 (changes requested):
This pretty much looks good to me; just a couple final questions about coverage and fallback.

src/lib/utilities-for-location.ts
oyamauchi: I wondered about how to handle the possibility of not showing a user's actual utility in my notes doc. Regardless of zip->utility mapping errors, we know there are utilities that even this dataset doesn't contain (i.e. there are several Vermont munis that the Vermont PUC lists that aren't in this dataset, as noted in the notes doc).

I think I would be satisfied, at least in the short term, with including an "Other" item in the frontend's utility selector, but not having a way for them to tell us what their utility is, since that would be a nontrivial design change. And maybe we can monitor for "Other" responses and see what zips they're in, and investigate from there?

We can include checking the utility options in the per-state QA process. I'll add that to the launch checklist.

scripts/generate-utility-data.ts
```diff
@@ -0,0 +1,250 @@
...
+import _ from 'lodash';
+import fetch from 'make-fetch-happen';
+import minimist from 'minimist';
+import path from 'path';
+import { promisify } from 'util';
+import xlsx from 'xlsx';
+import { createAuthorityName } from './lib/authority-and-program-updater';
+
+/**
+ * Utility codes (number) or utility names (string) to exclude from
```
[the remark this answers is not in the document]
oyamauchi: I thought there was a case where I had to do a name replacement on a utility that had `Not Available` as the utility code, but now I can't find it. There are a lot of Not Availables, so we'll likely need this functionality in the future, but I'll take it out for now as it's not needed.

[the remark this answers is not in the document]
datu925: You can also leave it in with a comment on when it would be needed in the future; it just wasn't apparent.

datu925 (approved):

PR #287, opened 2024-02-12
Remove savings from API v1

oyamauchi:
## Description

These aren't being used in the frontend, and we've long known they're
a bit iffy in various ways. v0 is untouched.

Task to remove from API: https://app.asana.com/0/1204738794846444/1206540986137853

Old task to remove from frontend, where we discussed some of the problems:
https://app.asana.com/0/1205057814651000/1205608789805592/f

## Test Plan

`yarn test`, look at fixture changes.

### DIFF
--- a/src/lib/incentives-calculation.ts
+++ b/src/lib/incentives-calculation.ts
@@ -47,6 +47,7 @@ export type CalculatedIncentive = (IRAIncentive | StateIncentive) & {
  */
 type CalculatedIncentives = Omit<APICalculatorResponse, 'incentives'> & {
   incentives: CalculatedIncentive[];
+  /** Only used in v0 for now */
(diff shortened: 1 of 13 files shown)


src/lib/incentives-calculation.ts
```diff
@@ -47,6 +47,7 @@ export type CalculatedIncentive = (IRAIncentive | StateIncentive) & {
  */
 type CalculatedIncentives = Omit<APICalculatorResponse, 'incentives'> & {
   incentives: CalculatedIncentive[];
+  savings: APISavings;
```
datu925, line 50: Optional: add comment saying this is for v0/not used by v1.

datu925 (approved):

PR #289, opened 2024-02-13
Remove type from v1 schema.

datu925:
Switching from `.type` to `.payment_methods[0]` in various backend calls is necessary since our CalculatedIncentive type no longer reliably has a `type` property. We could split up that type but I figured it was just easier to start using `payment_methods[0]` in the calculations anyway.

To make that safer, I added minItems requirements to the JSON schema def for payment_methods and owner_status; now those arrays can't be empty.

Tested: submitted an empty payment_methods array in one of the incentives and confirmed we got a test failure.

### DIFF
--- a/data/AZ/incentives.json
+++ b/data/AZ/incentives.json
@@ -3,7 +3,6 @@
     "id": "AZ-1",
     "authority_type": "utility",
     "authority": "az-mohave-electric-cooperative",
-    "type": "rebate",
     "payment_methods": [
       "rebate"
     ],
@@ -22,7 +21,6 @@
     "id": "AZ-2",
     "authority_type": "utility",
     "authority": "az-mohave-electric-cooperative",
-    "type": "rebate",
     "item": "heat_pump_air_conditioner_heater",
     "payment_methods": [
       "rebate"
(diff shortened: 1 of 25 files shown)


oyamauchi (approved):
It feels sketchy to be putting `payment_methods[0]` and according special significance to the first payment method in the list. I think this is OK because most instances of this are either on the federal incentives, which we know only have one method, or are for calculating savings totals, which don't get shown for state incentives now.

It may be worth noting, either in spreadsheet guidelines or in conversion scripts, the fact that if an incentive has a "primary" payment method, it should be first in the list?

datu925:
> It feels sketchy to be putting `payment_methods[0]` and according special significance to the first payment method in the list. I think this is OK because most instances of this are either on the federal incentives, which we know only have one method, or are for calculating savings totals, which don't get shown for state incentives now.
> 
> It may be worth noting, either in spreadsheet guidelines or in conversion scripts, the fact that if an incentive has a "primary" payment method, it should be first in the list?

I added a note on the order to the spreadsheet data models. I do think the root cause of this is that we haven't decided how we should calculate these things in a world where there can actually be multiple payment methods. Filed https://app.asana.com/0/1205271193902672/1206614782434228 to track.

PR #290, opened 2024-02-15
Add ability to explicitly ignore columns in spreadsheet to json flow.

datu925:
I've noticed that people will add exploratory columns to spreadsheets that aren't part of the main schema (example: Tom C added character and letter count columns to CO). We want to be able to ignore these while not opening up the possibility of *accidentally* missing out on columns due to typos.

Testing: ran a couple tests and the columns were ignored as expected in the resulting JSON.


mmezher:
IIUC this change isn't necessary since we likely want to restrict the columns allowed in the incentives sheets - right?

datu925:
> IIUC this change isn't necessary since we likely want to restrict the columns allowed in the incentives sheets - right?

Sorry, I was chatting with Tom about it and forgot to close the loop. I *think* we can avoid this change, but I'm still waiting to see how some other changes land before declaring victory.

Let me leave it up for a bit, but I'll take you off approver and add you again if it seems like we will need it after all.

datu925:
I think we're not going to need this; will reopen if I'm wrong.

PR #292, opened 2024-02-15
Add format-converter to bridge from csv-parse to our collected schema.

datu925:
This is not integrated yet – that is an upcoming PR.

I was tempted to call these methods csvToJson, but really csv-parse has already converted csv to json; it's just a flat object that we can't do interesting things with. These methods "inflate" that object and optionally validate it.

I noticed an error or two in the previous schema as testing, so those are fixed here.

### DIFF
--- a/scripts/lib/format-converter.ts
+++ b/scripts/lib/format-converter.ts
@@ -1,0 +1,90 @@
+import Ajv from 'ajv';
+import {
+  COLLECTED_DATA_SCHEMA,
+  CollectedFields,
+} from '../../src/data/state_incentives';
+import { LOCALIZABLE_STRING_SCHEMA } from '../../src/data/types/localizable-string';
+
+const ajv = new Ajv({ allErrors: true, coerceTypes: 'array' });
+
+// Could be generated from CollectedFields, filtering to array types.
+const ARRAY_FIELDS = ['data_urls', 'payment_methods', 'owner_status'];
+
+const validate = ajv
+  .addSchema(LOCALIZABLE_STRING_SCHEMA)
+  .compile(COLLECTED_DATA_SCHEMA);
+
+// Call this function to perform validation on the json.
+// This returns a tuple of valid and invalid records.
+// To convert without validation, call csvToJson directly.
+export function csvToJsonValidate(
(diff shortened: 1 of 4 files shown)


scripts/lib/format-converter.ts
```diff
@@ -0,0 +1,100 @@
...
+    for (const columnName in row) {
+      let val = row[columnName];
+      if (val === '') continue;
+      if (arrayCols.includes(columnName)) {
+        // Assume comma-delimited array and split into components.
+        val = val.split(',').map((s: string) => s.trim());
+      }
+
+      const chunks = columnName.split('.');
+      let dest = data;
```
mmezher, line 83: may be a typescript syntax best practice, but why create a new variable that is equivalent to data for modification, rather than just using the data {} var itself here?

test/scripts/format-converter.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    data_urls: 'https://takechargeva.com/programs/for-your-home',
+    authority_type: 'utility',
+    authority_name: 'Appalachian Power',
+    program_title: 'Take Charge Virginia Efficient Products Program',
+    program_url:
+      'https://takechargeva.com/programs/for-your-home/efficient-products-program-appliances',
+    item: 'heat_pump_clothes_dryer',
+    item_if_selected_other: '',
+    'short_description.en':
+      'Receive up to $50 rebate for an Energy Start certified electric ventless or vented clothes dryer from an approved retailer.',
```
mmezher, line 68: may need to change this if cp/paste - Energy Star?

mmezher (approved):
A few comments but nothing that is immediately blocking.

scripts/lib/format-converter.ts
datu925: I added some comments and renamed some variables to better explain what's going on. Basically you need one variable pointing to the root output object and one "cursor" variable pointing to the object or sub-object that is where you're prepared to write. That second variable always starts at the root for each key, but may advance into sub-objects. I don't see a way to do it with a single variable, but feel free to edit later if I'm missing something.

PR #293, opened 2024-02-15
Validate spreadsheets first in spreadsheet-to-json flow.

datu925:
This incorporates the format-converter and collected schema into the spreadsheet-to-json flow. Now we first validate that the collected data is valid under our spreadsheet schema, then take the collected data and turn it into a refined StateIncentive. This may surface some more nits in the short-term while we clean up the spreadsheets, but in the long-term is better since we can keep those spreadsheets healthy from the get-go.

It simplifies the `refineCollectedData` method since we can assumed we're working with valid collected data from the beginning. However, it did require rewriting a bit of the LowIncomeThreshold part since we could no longer just assign a string.

After this PR, I'll also pull out a stand-alone version of the CollectedFields validation that can be run on our spreadsheets on a periodic basis so we can be aware of any spreadsheet validity regressions.

### DIFF
--- a/scripts/lib/spreadsheet-standardizer.ts
+++ b/scripts/lib/spreadsheet-standardizer.ts
@@ -109,6 +109,8 @@ export class SpreadsheetStandardizer {
     record: CollectedFields,
   ): Partial<StateIncentive> {
     const authorityName = createAuthorityName(state, record.authority_name);
+    // Pass-through fields are those in CollectedFields that appear in the
+    // refined StateIncentive verbatim. Everything else needs some processing.
     const output: Partial<StateIncentive> = _.pick(record, PASS_THROUGH_FIELDS);
     output.authority = authorityName;
     output.program = createProgramName(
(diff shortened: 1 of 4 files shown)


scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -151,43 +145,27 @@ export class SpreadsheetStandardizer {
...
-      input['amount.minimum'] !== ''
-    ) {
-      amount.minimum = +cleanDollars(input['amount.minimum']);
-    }
-    if (
-      input['amount.maximum'] !== undefined &&
-      input['amount.maximum'] !== ''
-    ) {
-      amount.maximum = +cleanDollars(input['amount.maximum']);
+    if (authority in this.lowIncomeThresholds![state]) {
```
mmezher, line 164: question: what does the syntax here mean: this.lowIncomeThresholds![state] vs this.lowIncomeThresholds[state]?

mmezher (approved):

datu925:
> LGTM - had some trouble reading how the passthrough data worked and then some bits were manipulated, but I think overall it makes sense (and might just need some documentation).

Added a couple comments to hopefully make the pass-through data clearer.

PR #299, opened 2024-02-20
Add illinois incentives in JSON

matelau:
Add illinois incentives in JSON

Translation request submitted Asana [Ticket](https://app.asana.com/0/1205225542800756/1206649953550848/f).


data/low_income_thresholds.json
```diff
@@ -154,6 +154,21 @@
       }
     }
   },
+  "IL": {
+    "default": {
+      "source_url": "https://dceo.illinois.gov/communityservices/homeweatherization.html",
+      "thresholds": {
```
matelau, line 160: I wasn't sure if this was still valuable to keep if no incentives had defined income restrictions, but I wanted to check. Slack thread: https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708466841224109

data/IL/incentives.json
```diff
@@ -0,0 +1,769 @@
...
+    "short_description": {
+      "en": "$750 rebate for replacing existing geothermal with new geothermal system with an electric backup."
+    }
+  },
+  {
+    "id": "IL-7",
+    "authority_type": "utility",
+    "authority": "il-corn-belt-energy-corporation",
+    "payment_methods": [
+      "unknown"
```
matelau, line 138: Outstanding question if this method should be unknown. See Slack thread: https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708466841224109

matelau: Ticket created here: https://app.asana.com/0/1206629812394927/1206629812396080/f

I'm not sure who would be expecting to manage these tickets though? Maybe move back into an incentive board after the swarm is over? @veekas

data/authorities.json
```diff
@@ -183,6 +183,27 @@
       }
     }
   },
+  "IL": {
+    "utility": {
```
oyamauchi, line 187: I just landed #286, which means this will need some updates: the utility IDs need to be part of that Energy Star dataset. You can see the available IDs by doing this:

- Run `yarn build`
- `sqlite3 incentives-api.db`
- `select utility_id from utilities where utility_id like 'il-%';`

In particular, you'll have to change:

- `il-corn-belt-energy-corporation` -> `il-corn-belt-energy`
- `il-mid-american-energy-company` -> `il-mid-american-energy`

data/authorities.json
```diff
@@ -183,6 +183,27 @@
...
+        "name": "Commonwealth Edison"
+      },
+      "il-corn-belt-energy-corporation": {
+        "name": "Corn Belt Energy Corporation"
+      },
+      "il-jo-carroll-energy-cooperative": {
+        "name": "Jo-Carroll Energy Cooperative"
+      },
+      "il-mid-american-energy-company": {
+        "name": "MidAmerican Energy Company       "
```
oyamauchi, line 198: Remove trailing spaces

data/IL/incentives.json
```diff
@@ -0,0 +1,1110 @@
...
+      "maximum": 15000
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Up to $15,000 for energy-related weatherization and repair work."
+    },
+    "low_income": "il-state-of-illinois"
```
oyamauchi, line 783: Does it not work if you put `default` here?

data/low_income_thresholds.json
```diff
@@ -154,6 +154,34 @@
...
+        "2": 39440,
+        "3": 49720,
+        "4": 60000,
+        "5": 70280,
+        "6": 80560,
+        "7": 90840,
+        "8": 93297
+      }
+    },
+    "il-state-of-illinois": {
```
oyamauchi, line 171: Since they're the same, I would keep `default` and delete this one, to stay with the convention in other states.

data/authorities.json
datu925: The spreadsheet-to-json script can eventually handle this but for now, you may want to amend the above to:
`select * from utilities where utility_id like 'il-%';` so it shows both the IDs and the display name.

Then change the spreadsheet so the Authority name is the same as the correct display name. spreadsheet-to-json then *should* come up with the correct ID on its own.

In the long-term we'll have validation on the spreadsheet so that you can only choose valid Authority names (or IDs) in the first place, but this change was landed a few hours ago so it'll be a bit awkward temporarily.

data/IL/incentives.json
datu925: There's a thread on this here: https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708549761480869

We can discuss on that thread, but switching these to default if there is only one program is fine for now.

matelau: I couldn't actually get this to load the default without code changes [here](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/main/scripts/lib/spreadsheet-standardizer.ts#L159-L172). I couldn't find the logic elsewhere that would fallback to the default even though it seems like that happens elsewhere like NY. I could use some guidance unless we are okay moving forward with the change I suggested in the thread.

data/authorities.json
matelau: How do we rectify the state level weatherization incentive IL-35, which the script is currently identifying as `il-state-of-illinois` I didn't see an equivalent "authority" in the sqllite query `select * from utilities where utility_id like 'il-%';` ?

matelau: done

data/IL/incentives.json
matelau: I guess the problem is that the script we've been running `node build/scripts/incentive-spreadsheet-to-json.js` will clobber any manual edits we make to this file. So, even if I manually set this to `default` it gets overwritten on the next run. Perhaps that isn't a big deal now that this is generated?

data/IL/incentives.json
datu925: Just flagging that this will eventually block IL from being moved to a launched state due to https://github.com/rewiringamerica/api.rewiringamerica.org/blob/1ae3463106091755ba23ada194bda49e7b5833d4/test/data/schemas.test.ts#L211.

That's fine for now, but that's something I put in place to ensure we don't start serving these before we've figure out better answers.

data/IL/incentives.json
```diff
@@ -0,0 +1,1110 @@
...
+    "program": "il_cornBeltEnergyCorporation_rebateProgram",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 2000
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Up to $2,000 rebate for geothermal heat pumps."
```
datu925, line 166: Some of these descriptions are a little too terse, IMO. Feel free to look at other states for examples and push back if I'm wrong, but in general we like to include some details about important restricts on e.g. SEER, max/min size, etc. if they are available (if not, don't worry about it).

datu925: Our maximum is 150 – you don't have to hit that, but I'd review anything below ~50 or so.

data/IL/incentives.json
```diff
@@ -0,0 +1,1110 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 225,
+      "maximum": 225
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Up to $225 per unit."
```
datu925, line 1105: Definitely too short – needs to mention the kind of appliance.

data/IL/incentives.json
```diff
@@ -0,0 +1,1110 @@
...
+    "end_date": 2024
+  },
+  {
+    "id": "IL-43",
+    "authority_type": "utility",
+    "authority": "il-jo-carroll-energy-cooperative",
+    "payment_methods": [
+      "account_credit"
+    ],
+    "item": "other",
```
datu925, line 954: These have an item now – it's non_heat_pump_water_heater, so you should use that instead of other. You may have to edit or ignore the data validations in Google sheets. Same for some of the other `other`s.

datu925 (changes requested):

data/IL/incentives.json
matelau: Went through and gave the short strings some more love.

data/IL/incentives.json
matelau: updated

data/IL/incentives.json
matelau: Where is the source of truth for these? Do we have one for energy audits as well?

matelau: found it https://github.com/rewiringamerica/api.rewiringamerica.org/blob/main/src/data/types/items.ts

matelau: @datu925 the remaining entities marked other should just be energy audits. It didn't look like we had a better classification for those yet.

datu925: > @datu925 the remaining entities marked other should just be energy audits. It didn't look like we had a better classification for those yet.

For better or for worse, it looks like we lump energy audits under weatherization (see https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708640580381249). Let's switch to that for now.

data/IL/incentives.json
matelau: I tried my luck by reaching out to the number on the [website](https://www.cornbeltenergy.com/geocents) 800-879-0339 to get more information. The representative took my contact info, but we will see how fruitful this turns out to be 🤞🏼.

datu925 (approved):
One issue with the incentive relationships, but otherwise looks good.

data/IL/incentive_relationships.json
```diff
@@ -0,0 +1,5 @@
+{
+  "prerequisites": {
+    "CT-40": "IL-50"
```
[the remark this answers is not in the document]
matelau: oppps, thanks for catching. I'll update.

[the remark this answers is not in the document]
matelau: done

PR #300, opened 2024-02-21
RI 2024 updates: adds e-bike rebate, add status and start and end date for some programs

rpartridge:
Asana ticket: https://app.asana.com/0/1204738794846444/1206378369410355/f
Google sheet for RI: https://docs.google.com/spreadsheets/d/1Ik9iN2jDyD_ON2WdL24pmWUo6vps28zWiHKyFfZG334/edit#gid=30198531

## Overview
This ticket was to 1) update the RI spreadsheet linked above and 2) add any relevant updates to the API

## Changes
1. Adds rebate for RI-13 (RI OER ebike)
2. Adds rebate for RI-43 (RI OER ebike, income qualified)
3. Adds exclusion file for RI so that users are ineligible for RI-13 if they qualify for RI-43
4. New program type for OER ebike
5. Updates fixtures for tests
6. Adds start_date, end_date for RI-20, RI-22. The ticket added these fields for a bunch of incentives in the spreadsheet, but these were the only 3 already in the calculator.
7. Removes RI-19 which is paused. We don't have support for `program_status` right now, so instead of updating the program status, I just removed this item (slack discussion [here](https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708969866417359))

Note that some updates are not included. I only updated the rows that currently exist in the API. See the RI google sheet columns Calc API Developer Notes for next steps on additional columns


data/RI/incentives.json
```diff
@@ -245,6 +245,32 @@
...
+      "maximum": 350
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "low_income": "default",
+    "short_description": {
+      "en": "30% up to $350 back on the final purchase of an e-bike or e-cargo bike. For income-qualified residents, 75% up to $750 is available.",
+      "es": ""
```
rpartridge, line 269: TODO

data/RI/incentives.json
```diff
@@ -245,6 +245,32 @@
...
     }
   },
+  {
+    "id": "RI-13",
+    "authority_type": "state",
+    "authority": "ri-oer",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "ebike",
```
rpartridge, line 255: I added this to the API because of Tom's comment here: https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708462830835669?thread_ts=1708461014.290839&cid=C04E4THEGMT

but the tests are failing with `launched states do not have any values that we don't support in the frontend > ebike not a launched technology
  in RI-13 and cannot be used in state: RI`. is this expected? should I update ebikes to be a launched item? https://github.com/rewiringamerica/api.rewiringamerica.org/blob/main/src/data/types/items.ts#L24

data/RI/incentives.json
```diff
@@ -245,6 +245,32 @@
...
+      "number": 0.3,
+      "maximum": 350
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "low_income": "default",
+    "short_description": {
+      "en": "30% up to $350 back on the final purchase of an e-bike or e-cargo bike. For income-qualified residents, 75% up to $750 is available.",
```
oyamauchi, line 268: In this case we'd rather have the income-qualified version expressed as a separate incentive, and have an `exclusion` relationship between them, expressing "if you claim the low-income version, you can't also claim the non-restricted version".

You can look at VT-55 and VT-56 as an example, but the steps are basically:

- Add the low-income version as its own incentive. (I think you can say it's under the same program.)

- Add a separate file `data/RI/incentive_relationships.json`:
  ```json
  {
    "exclusions": {
      "RI-XX": ["RI-YY"]
    }
  }
  ```
  The low-income incentive's ID should be the key in the map, which expresses that it's the one that "blocks out" the other.

- Add code to read that file in `state_incentive_relationships.ts`.

And ideally it should be two separate rows in the spreadsheet too, to ease eventually using the script to do the translation.

data/RI/incentives.json
oyamauchi: Not sure about this one -- I don't have a sense of the full implications of making an item "launched" (I don't even remember seeing this code change go by). I think this is a question for Slack.

oyamauchi (changes requested):
There's also the PUD smart thermostat incentives. Smart thermostats are a "beta" item, though, so whether you can add them now depends on the answer to the ebikes question.

It looks like the Google sheet now reflects the updates from the emailed Excel file from RI Energy; is that right?

data/RI/incentives.json
rpartridge: followed up in [slack](https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708539293768709?thread_ts=1708461014.290839&cid=C04E4THEGMT) with @datu925 . sounds like the beta vs launched distinction was to ensure that all launched items have a `more_info` url, but we want to move away from that (Dan is handling removing that requirement). we also want to ensure that all launched items have Spanish translations.

based on that this PR is blocked by:
1. Spanish translations for the 2 new ebike rebates ([ticket created for CCCE here](https://app.asana.com/0/1205381742684374/1206659011068825/f))
2. Dan's updates to enable us to "launch" ebikes without the `more_info` url

data/RI/incentives.json
rpartridge: updated!

data/RI/incentives.json
```diff
@@ -245,6 +245,54 @@
...
+    "amount": {
+      "type": "percent",
+      "number": 0.3,
+      "maximum": 350
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "low_income": "default",
```
rpartridge, line 266: confirmed that the income restrictions on the [website](https://drive.ri.gov/erika-niedowski-memorial-electric-bicycle-rebate-program) are the same as the [default ones for RI](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/main/data/low_income_thresholds.json#L173-L187)

data/RI/incentives.json
datu925: https://github.com/rewiringamerica/api.rewiringamerica.org/pull/304 landing should unblock #2.

data/RI/incentives.json
```diff
@@ -245,6 +245,54 @@
       "es": "Un reembolso adicional de $1,500 por un vehículo eléctrico usado para los calificadores de LIHEAP. Por orden de llegada."
     }
   },
+  {
+    "id": "RI-13",
```
rpartridge, line 249: I tried to add fields `program_status` and `other_restrictions`, but they were causing test failures, I think because they're not included [here](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/main/src/data/state_incentives.ts#L171-L174). I ended up removing those fields and it sounds like from slack, program_status isn't used by the calculator anyway (is other_restrictions?) so maybe this is fine as-is.

oyamauchi (approved):
LGTM

datu925 (approved):
Looks good to me. Your PR description might be slightly out of date (re adding fields like program_description).

You might also want to mention that some of the RI records still aren't making it into JSON (RI-36 through RI-42) and summarize why.

PR #301, opened 2024-02-21
Add spreadsheet health test and register CO and AZ as passing states.

datu925:
This test validates that the spreadsheets match our collected data schema. It is skipped right now, but likely next step is to add a workflow that sets an environment variable so we can run it only in those contexts.

CO and AZ are the only tests currently passing :( but I will circle back to the others later.


mmezher (approved):

test/data/spreadsheets-health.test.ts
```diff
@@ -0,0 +1,63 @@
+import { parse } from 'csv-parse/sync';
+import fetch from 'make-fetch-happen';
```
mmezher, line 2: An aside: I can't believe this is a real import.

PR #302, opened 2024-02-21
Adding Spanish descriptions for CT

mariabajzek:
I also increased the character limit for Spanish descriptions in the schema test from 160 to 200. On Slack, we had discussed increasing to 180 based on the idea that on average Spanish is 20% longer than English. However, even 200 still doesn't cover a few of the longer descriptions. I left those out for now but they should be added back when we have a better way to display longer descriptions.

[Here's](https://docs.google.com/spreadsheets/d/13mJyltzLt4KTDuuL14r71HqKJU5c9f6bYJtGLotgkVI/edit?pli=1#gid=995688950) the spreadsheet with the descriptions for reference.


data/CT/incentives.json
```diff
@@ -153,7 +155,8 @@
       "homeowner"
     ],
     "short_description": {
-      "en": "Free energy assessment, with possible weatherization followups."
+      "en": "Free energy assessment, with possible weatherization followups",
```
mmezher, line 158: nit: consistently, these short descriptions have punctuation. Add period to end?

mariabajzek: Fixed in the spreadsheet and in the JSON.

mariabajzek:
Chatted with Monique offline, and we decided that it's worse for there to be no Spanish description for a few of these than for a few of them to be a bit longer, so I increased the limit and added them back. Will bring it up in the standup later to see what action we want to take about it.

PR #304, opened 2024-02-22
Remove requirement that items have locale urls and remove beta vs launched item distinction.

datu925:
We've decided to decouple the API and frontend and so in general, we want to bias toward making things available in the API sooner rather than later. We've required items to have more_info urls for all locales, but we aren't going to have these for more "minor" items going forward – or at least, we don't want to block getting them in the API on that.

In order to make that safe for v0, I've added a static test that all IRA incentives do have urls.

As a consequence, the beta vs launched distinction no longer serves a purpose so I'm removing it. That does mean that adding a new item does require an English and Spanish localization, but that's a small enough amount of work that I think it's fine. These translations are Dena-approved.

The initial trigger for this is https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1708537478303219?thread_ts=1708461014.290839&cid=C04E4THEGMT.


locales/es.json
```diff
@@ -17,19 +17,28 @@
...
     "heat_pump_water_heater": "Calentador de agua con bomba de calor",
     "new_electric_vehicle": "Nuevo vehículo eléctrico",
+    "non_heat_pump_clothes_dryer": "Secadora de ropa eléctrica",
+    "non_heat_pump_water_heater": "Calentador de agua eléctrico",
+    "other": "Otro",
     "rooftop_solar_installation": "Instalación solar en el techo",
+    "smart_thermostat": "Termostato inteligente",
     "used_electric_vehicle": "Vehículo eléctrico usado",
-    "weatherization": "Climatización"
+    "weatherization": "Climatización",
```
oyamauchi, line 40: This is different from what the embed frontend has. It refers to HVAC as `sistemas de climatización` and weatherization as `impermeabilización`. I'm OK with leaving it for now, since I'm not even sure if we have RA-wide standards for this.

oyamauchi (approved):

PR #309, opened 2024-02-26
Add WI incentives

hblumberg:
Note that this work did NOT include a data refresh. Carden has reached out to Focus on Energy to see if they're able to refresh the data.

## Links
- _Asana_ https://app.asana.com/0/1206629812394927/1206662027052982/f

## Checklist

## Test Plan
* Ensure `node build/scripts/incentive-spreadsheet-to-json.js --strict WI` runs without error
* Ensure `yarn test` runs without error


data/low_income_thresholds.json
```diff
@@ -320,5 +320,33 @@
...
+        "2": 75750,
+        "3": 85200,
+        "4": 94650,
+        "5": 102250,
+        "6": 109800,
+        "7": 117400,
+        "8": 124950
+      }
+    },
+    "wi-focus-on-energy": {
```
datu925, line 338: This is a bug/awkward spot with the tooling right now where it is much easier/more automated to define a dummy default that is never used than to use the default.

For that reason, I'm fine leaving this as is, though other reviewers may ask you to remove the wi-focus-on-energy entry.

scripts/incentive-spreadsheet-registry.ts
```diff
@@ -63,4 +63,13 @@ export const FILES: { [ident: string]: IncentiveFile } = {
...
     esHeader: 'Program Description (Spanish)',
   },
+  WI: {
+    filepath: 'data/WI/incentives.json',
+    sheetUrl:
+      'https://docs.google.com/spreadsheets/d/e/2PACX-1vSxNnovJ_viGErGJIkLL8uyXNx6QWKSgLZlntceubHXeYbb_s3Ksjxgg6ZSU8MiayjtmaMUpiHC1ue9/pub?gid=995688950&single=true&output=csv',
+    headerRowNumber: 2,
+    idHeader: 'ID',
+    enHeader: 'Program Description (style guide)',
+    runSpreadsheetHealthCheck: true,
```
datu925, line 73: In order to confirm this actually passes the test, you need to temporarily flip the `skip` boolean in `test/data/spreadsheets-health.test.ts` to false. The test is skipped by default.

If it doesn't pass, that's fine, just remove this line from the spreadsheet registry. This test is evolving fast anyway so I wouldn't invest a lot of time making it pass.

Either way, revert the skip boolean when you're done.

data/WI/incentives.json
```diff
@@ -0,0 +1,667 @@
...
+    "end_date": 2023,
+    "bonus_available": true,
+    "low_income": "wi-focus-on-energy"
+  },
+  {
+    "id": "WI-8",
+    "authority_type": "state",
+    "authority": "wi-focus-on-energy",
+    "payment_methods": [
+      "unknown"
```
datu925, line 190: This pretty clearly seems like an instant discount/point-of-sale rebate from the website.

data/WI/incentives.json
```diff
@@ -0,0 +1,667 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 120,
+      "minimum": 120
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Instant discounts for natural gas tankless water heaters start at $120 through a Trade Ally contractor."
```
datu925, line 226: We absolutely do not want to show incentives for gas appliances :)

You can delete this from the source spreadsheet. Same for W12-W20.

data/WI/incentives.json
```diff
@@ -0,0 +1,667 @@
+[
+  {
+    "id": "WI-1",
```
datu925, line 3: The descriptions on WI-1 through WI-7 are a little short IMO. We usually like to include more data than just the price and item when it's available.

My rule of thumb is that you don't have always try to hit the 150 character limit, but if you're under 50 characters, it should only be because the website didn't give any other information. I do think there are other potentially interesting program requirements in this case.

datu925 (changes requested):

data/WI/incentives.json
hblumberg: Good catch, thank you!

scripts/update-fixtures.sh
```diff
@@ -164,3 +164,16 @@ curl \
...
+curl \
+  "http://localhost:3000/api/v1/calculator?\
+location%5Bzip%5D=53703\
+&include_beta_states=true\
+&owner_status=homeowner\
+&household_income=50000\
+&tax_filing=joint\
+&household_size=1\
+&authority_types=state\
+&authority=wi-focus-on-energy" \
```
veekas, line 178: curious why you added the "authority" filter? I don't see it in other fixtures (and I'm wondering if I should add it to mine)

veekas (approved):

data/WI/incentives.json
hblumberg: I elaborated on a few of the descriptions, but I couldn't find anything else to say about attic or wall insulation.

scripts/update-fixtures.sh
hblumberg: Oh good question! I copied IL. But it's the only authority so it doesn't matter whether or not I include it. (I removed it for consistency!)

data/WI/incentives.json
datu925: FYI, there's a lot of extra info that might be relevant in the Program Requirements tab on that website (under Rebate Details).

The hierarchy of what to include is more art than science, but it's nice to have some more detail to offer the user before they need to go hunting themselves.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+    "authority": "wi-focus-on-energy",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "weatherization",
+    "program": "wi_focusOnEnergy_insulation&AirSealingRebates",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 75,
+      "minimum": 75,
```
datu925, line 118: You don't need the minimums and maximums if the dollar amount is flat and the same. Delete these from the spreadsheet in those cases.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Up to $1,125 for ENERGY STAR Qualified Air Sealing. Must be completed with Trade Ally contractor installed attic or wall insulation improvements."
+    },
+    "start_date": 2023,
+    "end_date": 2023,
+    "low_income": "wi-focus-on-energy"
```
datu925, line 26: For these first 3, it seems like there are two versions of each incentive – one income qualified and one not. I would say the right thing to do is to split those rows (in the spreadsheet), specify the more exact amounts, and update the descriptions to indicate whether they are low-income qualified or not. Then only one of each pair will get the low-income tag.

You might need to redo the IDs as a result, but that's fine, they aren't checked-in yet anyway.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$450 for Wall Insulation."
+    },
+    "start_date": 2023,
+    "end_date": 2023,
+    "low_income": "wi-focus-on-energy"
```
datu925, line 104: If I'm understanding correctly, from this point on, the incentives don't actually have low-income restrictions, so this shouldn't be here. If the script is putting this in, it's because those rows have n/a in the low-income column. Delete the n/a's and rerun and they should only appear in the rows where low-income is actually populated.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+      "number": 50,
+      "minimum": 50,
+      "maximum": 50
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$50 for Smart Thermostats. Available as instant discounts in the online marketplace or as rebates."
```
datu925, line 178: If this is true, then the "Rebate Type" column should have "Point of Sale rebate" populated as well. You may need to remove the data validations on the spreadsheet in order to be able to put both.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+      "number": 500,
+      "minimum": 500,
+      "maximum": 500
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Rural residential customers can receive an additional $500 for installing a system."
```
datu925, line 228: I think this description doesn't make much sense as stated. What kind of system? Additional over what? Is there some other incentive that this is a bonus on top of?

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+      "number": 500,
+      "minimum": 500,
+      "maximum": 500
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$500 for Cold Climate Air Source Heat Pump (existing primary heating fuel: propane, oil, etc.)."
```
datu925, line 305: Same as above.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+    ],
+    "short_description": {
+      "en": "$200 cash back for qualified self-installed attic insulation and air sealing."
+    },
+    "start_date": 2023,
+    "end_date": 2023,
+    "low_income": "wi-focus-on-energy"
+  },
+  {
+    "id": "WI-21",
```
datu925, line 261: If I'm understanding correctly, W21 and W22 seem like out of date duplicates of W23 and W24. Or at least, I can't find anywhere on that page where W21/22 exist.

If so, you can keep the 2023 ones, but do update the 2024 ones to have the correct start/end dates so it's more obvious that these are dupes.

data/WI/incentives.json
```diff
@@ -0,0 +1,410 @@
...
+      "number": 1300,
+      "minimum": 1300,
+      "maximum": 1300
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$1,300 for Cold Climate Air Source Heat Pump (existing primary heating fuel: natural gas or electric)."
```
datu925, line 280: For all of these, this phrasing seems a bit more succinct: $1,300 for Cold Climate Air Source Heat Pump replacing natural gas or electric primary heating source.

datu925 (changes requested):
I know you aren't supposed to do a data refresh, but a lot of these comments are more about correcting/vetting the initial set of data vs refreshing for 2024, so I think they still make sense to do.

hblumberg:
> I know you aren't supposed to do a data refresh, but a lot of these comments are more about correcting/vetting the initial set of data vs refreshing for 2024, so I think they still make sense to do.

Got it, I think I misunderstood what a data refresh is! I'll make these changes!

data/WI/incentives.json
hblumberg: Huh, good catch! I'll remove W21 and W22

datu925 (approved):

PR #310, opened 2024-02-26
Add Nevada incentives

rpartridge:
Adds all incentives for Nevada. See spreadsheet: https://docs.google.com/spreadsheets/d/1w6c4lAlNGTKEuN_X00mAcK3Xs6NGM1aT5XFbhjd1CGo/edit#gid=995688950

Note that we were advised NOT to do a data refresh for NV. I ended up doing a partial data refresh anyway because a bunch of validations were failing, mostly:
- Getting descriptions <= 150 chars
- Program name and URL were missing for most programs
- Owner type was missing for almost all rows
- Ensuring amount, amount_type, and units passed validations

However, I did not go through all the data comprehensively, and only cleaned up the descriptions when they were over 150 chars. I also didn't submit a Spanish translation ticket because once the data refresh happens, I'm sure the English translations will change quite a bit anyway.


data/authorities.json
```diff
@@ -204,6 +204,46 @@
...
+      "nv-rural-nevada-development-corporation": {
+        "name": "Rural Nevada Development Corporation"
+      },
+      "nv-nevada-rural-housing": {
+        "name": "Nevada Rural Housing"
+      },
+      "nv-help-of-southern-nevada": {
+        "name": "HELP of Southern Nevada"
+      }
+    }
```
rpartridge, line 245: these were marked as "other" in the spreadsheet because they're independent nonprofits. is this the right way to represent nonprofits?

data/low_income_thresholds.json
```diff
@@ -169,6 +169,34 @@
...
+        "1": 30120,
+        "2": 40880,
+        "3": 51640,
+        "4": 62400,
+        "5": 73160,
+        "6": 83920,
+        "7": 94680,
+        "8": 105440
+      }
+    },
```
rpartridge, line 185: this isn't really the default for NV, but it applies to the 4 nonprofits (CSA Reno, Rural Nevada Development Corporation, Nevada Rural Housing, HELP of Southern Nevada). rather than write out the same thresholds 4 times I set it as the default. if that's not the pattern we want to follow I can be more explicit

scripts/update-fixtures.sh
```diff
@@ -101,6 +101,20 @@ location%5Bzip%5D=60304\
...
+curl \
+  "http://localhost:3000/api/v1/calculator\
+?location\[zip\]=89108\
+&include_beta_states=true\
+&owner_status=homeowner\
+&household_income=40000\
+&tax_filing=joint\
+&household_size=1\
+&authority_types=utility\
+&authority_types=other\
```
rpartridge, line 114: including the "other" authority types because they have low income thresholds

test/data/schemas.test.ts
```diff
@@ -123,6 +126,10 @@ function isIncentiveAmountValid<T extends StateIncentive>(
       return amount.number >= 0 && amount.number <= 1 && !amount.unit;
     case 'dollars_per_unit':
       return amount.unit !== undefined;
+    case 'other':
+      return true;
+    case 'unknown':
+      return true;
```
rpartridge, line 132: there were a few weird cases that led to adding these.

1. NV-17 is a program for a [free clothes dryer](https://www.nvenergy.com/save-with-powershift/qualified-appliance-replacement). I marked this as "other". as far as I can tell on the website there isn't a monetary value, everywhere just says "free ENERGY STAR® electric clothes dryer"
2. NV-26 - program for a [free smart thermostat](https://www.nvenergy.com/save-with-powershift/smart-thermostat). I also marked this as "other"
3. NV-57 and NV-58 had no information about how much the rebate was actually for ([example application](https://www.rrelectric.com/wp-content/uploads/Residential-Insulation-Measures-Application.pdf)). it pretty much just says contact office to get more details... I marked these as "unknown".

let me know if this is the wrong approach

datu925: For free items, we express those as having amount type percent and the value is 1, representing 100% of costs (100% of eligible costs is what you can put in the Rebate Value column).

For NV-57 and NV-58, for similar cases in the past we've made the decision not to show those incentives at all in the JSON (see https://docs.google.com/document/d/18kkfkuz1Wn6_FYWid0U1v5VsJ8FUTELzzqhWJk1sbfM/edit, question #5).

I have seen that "unknown" is a handy value to have, at least for the spreadsheets, so I might add that later, but for now I don't think we want to support it and I don't think it should be valid for API data.

data/low_income_thresholds.json
datu925: Seems fine to me.

data/authorities.json
datu925: Hmm, these are tricky.

1. Even if we don't change much else, I don't think we should have an "other" category for AuthorityType. If we do need a separate category, I think we should add some kind of private/nonprofit option.

2. Most of these are eligibility-bounded by county (can't tell with the HELP one). This does seem like an example of authority and eligibility being largely unrelated. They're not County authorities, but they have County eligibility boundaries.

Going to defer to @oyamauchi on how to proceed.

oyamauchi: This is running up against the exact problem I'm trying to work on now. Here's the story from the beginning:

- Originally, I thought that the "authority" (i.e. the entity offering/administering the incentive) would be coterminous with the set of people who were eligible for the incentive based on location. I.e. state-level authorities would offer incentives to everyone in the state, utilities would offer incentives to their customers, city-level authorities would offer to their residents, etc. Notably, "state-level authority" didn't necessarily mean "part of the state government", it meant any authority who offered incentives to all state residents.
- This has turned out not to be the case. One problem we have in VT and CO is the existence of consortiums of utilities, which have a distinct identity and offer incentives to customers of any of several utilities.
- These nonprofits are another assumption-breaking case: their incentives are actually geographically restricted, by county. (E.g. [this page](https://rndcnv.org/weatherization-services-rural-nevada-development-corporation/) has a list of counties where the incentive is available.)
- So I'm thinking it's time to formally separate these concepts. I'm working on a change to do so for utilities; I think it'll be a new field on incentives that contains a list of utilities.
- Following on from _that_, I'm thinking the distinction between authority types in this file might become moot? The authority type/identity would stop serving as an eligibility criterion at all, and would instead just be a label for who administers the incentive.

So that's my thinking so far; sorry for the wall of text. (Your opinions on this stuff are welcome!) For this PR, I think what you have here is fine, with the understanding that it'll probably change soon.

data/NV/incentives.json
```diff
@@ -0,0 +1,1712 @@
...
+    "authority_type": "other",
+    "authority": "nv-rural-nevada-development-corporation",
+    "payment_methods": [
+      "pos_rebate"
+    ],
+    "item": "weatherization",
+    "program": "nv_ruralNevadaDevelopmentCorporation_weatherizationAssistanceProgram",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 12666
```
oyamauchi, line 1634: Where did this number come from? It'll show up in the frontend as "$12,666 off weatherization" which I think is not right. According to [the webpage](https://rndcnv.org/weatherization-services-rural-nevada-development-corporation/), it's just free weatherization assistance?

data/NV/incentives.json
```diff
@@ -0,0 +1,1712 @@
...
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "The Weatherization Assistance Program implements home energy conservation measures at no cost for residents below 200% of the federal poverty line."
+    },
+    "low_income": "default"
```
oyamauchi, line 1643: This doesn't seem right either; the webpage says 150% of the federal poverty line, and I don't see a reference to the state thresholds?

oyamauchi (changes requested):
Requesting changes for at least the free-stuff incentives; I'm also kinda concerned about those rural nonprofit weatherization incentives for various reasons (in comments)

test/data/schemas.test.ts
datu925: Didn't see Owen's comment before adding mine, but looks like we're generally aligned.

data/NV/incentives.json
rpartridge: I didn't do a data refresh so this is just what was already in the spreadsheet, not sure where it came from. Looking through their webpage/application though I don't see a monetary value referenced anywhere so to your point [here](https://github.com/rewiringamerica/api.rewiringamerica.org/pull/310/files#r1504591937) I'll go ahead and remove this incentive anyway

data/authorities.json
rpartridge: This is really helpful, thank you! One other consideration is that for at least some of these programs (like [NV-76](https://nvrural.org/programs/weatherization/)) the income qualifications also differ by county, in some counties it's 150% of the federal poverty line and in others it's 200%.

After looking at these 4 websites more closely though I actually don't see anywhere where it says what the value of the rebate is, the values in the spreadsheet seem to be out of date. Based on that I think I'm going to remove these anyway

rpartridge:
In the latest commit:
- Updated how the free items are represented (percent with value of 1)
- Removed all items with unknown values, which included all of the incentives available through the nonprofit programs

For the future is there a way in the spreadsheet to mark some rows to get skipped? Otherwise the "unknown" rows will keep getting repopulated every time we run the script and will need to be manually deleted. Or should we make a ticket to update the script to skip rows with unknown amounts?

datu925:
> In the latest commit:
> 
> * Updated how the free items are represented (percent with value of 1)
> * Removed all items with unknown values, which included all of the incentives available through the nonprofit programs
> 
> For the future is there a way in the spreadsheet to mark some rows to get skipped? Otherwise the "unknown" rows will keep getting repopulated every time we run the script and will need to be manually deleted. Or should we make a ticket to update the script to skip rows with unknown amounts?

This is actually on-deck – I am working on a general-purpose way to mark rows in the spreadsheet as "skip". It won't land until later this week – in the meantime, it's probably easiest to delete them manually after each run.

data/authorities.json
veekas: Just for visibility, a similar issue presented itself with DC, where a private entity was given the sole authority to offer energy rebates on behalf of the city. DC government confirmed that it would best be labeled "city". (One somewhat related note is that DC recently changed the law so that utilities can also offer rebates to their own customers.)

veekas (approved):
looks good enough to move forward to me!

oyamauchi (approved):

PR #311, opened 2024-02-27
Add smart thermostat incentives to RI

rpartridge:
# Changes

This PR adds the 2 new smart thermostat incentives available through PUD. I split this out from the original PR (which introduced new ebike incentives in RI) to reduce the risk releasing 2 beta items at once.

This shouldn't change any incentives available in the calculator since the calculator won't be requesting smart thermostat incentives.

All changes:
- 2 new smart thermostat incentives in `incentives.json`
- Based on the [program info](https://pud-ri.org/wp-content/uploads/2022/01/Residential-Energy-Audit-Weatherization-Incentives.pdf), added some incentive relationships:
    - exclusion between RI-41 and RI-42 (it specifies you can only get the $100 rebate OR the $25 one)
    - RI-4 is a prerequisite for both RI-41 and RI-42 because "Programmable thermostats will only qualify if recommended in the home energy audit" (RI-4 is the energy audit mentioned)

# Resources

- [PUD program rules](https://pud-ri.org/wp-content/uploads/2022/01/Residential-Energy-Audit-Weatherization-Incentives.pdf)
- [RI incentives spreadsheet](https://docs.google.com/spreadsheets/d/1Ik9iN2jDyD_ON2WdL24pmWUo6vps28zWiHKyFfZG334/edit#gid=533322917)
- [original ticket](https://app.asana.com/0/1206629812394927/1206378369410355/f)
- [new ticket](https://app.asana.com/0/1206629812394927/1206659708264490/f)
- [original PR](https://github.com/rewiringamerica/api.rewiringamerica.org/pull/300)

### DIFF
--- a/data/RI/incentive_relationships.json
+++ b/data/RI/incentive_relationships.json
@@ -2,6 +2,13 @@
   "exclusions": {
     "RI-43": [
       "RI-13"
+    ],
+    "RI-41": [
+      "RI-42"
     ]
+  },
+  "prerequesites": {
+    "RI-41": "RI-4",
+    "RI-42": "RI-4"
   }
 }
--- a/data/RI/incentives.json
+++ b/data/RI/incentives.json
@@ -544,5 +544,51 @@
       "en": "Depending on your income, 100% of the cost to install a heat pump water heater, including up to $3,000 towards related electric service upgrades.",
       "es": "Según sus ingresos, el costo total de instalar un calentador de agua con bomba de calor. Incluye hasta $3,000 para mejoras eléctricas relacionadas."
     }
+  },
+  {
+    "id": "RI-41",
+    "authority_type": "utility",
+    "authority": "ri-pascoag-utility-district",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "smart_thermostat",
+    "program": "ri_residentialEnergyAuditWeatherization",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 100,
+      "maximum": 100
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$100 back on programmable wireless connection enabled thermostat.",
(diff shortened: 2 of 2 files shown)


oyamauchi (changes requested):
I kinda think these make more sense as a single incentive, for up to $100 with the "$25 for non-wireless" part in the description. The mutual exclusion relationship makes sense, but it means nobody will ever see RI-42 -- their content is exactly the same except for amount, so they'll always have the same eligibility value (before relationships logic is applied). It seems weird to have an incentive that is guaranteed never to be returned as eligible.

One thing that would make this be sensible as two incentives is if we tracked wireless and non-wireless thermostats as separate `item` values, but I think that's too fine of a distinction and I doubt we'd ever do it.

Do you think all that makes sense? I could be convinced otherwise.

rpartridge:
> I kinda think these make more sense as a single incentive, for up to $100 with the "$25 for non-wireless" part in the description. The mutual exclusion relationship makes sense, but it means nobody will ever see RI-42 -- their content is exactly the same except for amount, so they'll always have the same eligibility value (before relationships logic is applied). It seems weird to have an incentive that is guaranteed never to be returned as eligible.
> 
> One thing that would make this be sensible as two incentives is if we tracked wireless and non-wireless thermostats as separate `item` values, but I think that's too fine of a distinction and I doubt we'd ever do it.
> 
> Do you think all that makes sense? I could be convinced otherwise.

Oh interesting, I didn't realize that we could never show anyone RI-42 in this case. So exclusions are only useful if the eligibility rules are different, like based on income?

I agree that separating out non-wireless thermostats into a different item type is probably not worth it.

I think combining makes sense then -- that's basically how we handle other incentive types, like heat pump incentives that offer different amounts for heat pumps of different efficiencies, but you can still only get 1 heat pump. Will update 👍

oyamauchi (approved):
Right, the exclusions we have now are just for (a) you can't claim both the low-income and non-LI versions of an incentive, and (b) incentives from different agencies that explicitly say, in their rules, you can't claim them together (there's a case of this in VT).

PR #312, opened 2024-02-28
Add DC Incentives

veekas:
Adds all incentives for DC: https://docs.google.com/spreadsheets/d/1ww57XfXAqkRL5t7bSqyoe3KpOxi-N-W_XBYbU43tNnc/edit#gid=995688950

Spreadsheet changes made:
- Getting descriptions <= 150 chars
- fixing `amount` values to be numbers instead of ranges
- Standardized program urls
- omit "other" programs

Other notes:
- Automated spreadsheet-to-JSON requires manual override for DC because of DOEE's WAP program needs to use the 60% AMI thresholds instead of the default 80%.
- This PR consolidates "Potomac Electric Power" to "Pepco" and removes "Washington Gas"


scripts/README.md
```diff
@@ -15,7 +15,7 @@ Use once the final English version is already available in a JSON file in `/data
 Usage:
 
 ```
-node build/scripts/generate-spanish-descriptions.js <state_id>
+npx ts-node build/scripts/generate-spanish-descriptions.ts <state_id>
```
veekas, line 18: I'm not sure if I'm missing something obvious here, but these are the scripts I had to run in order to run these TS files

scripts/lib/spreadsheet-mappings.ts
```diff
@@ -5,7 +5,7 @@ export type FlatAliasMap = { [index: string]: string };
 
 // Map from canonical column name to possible aliases in spreadsheets.
 export const FIELD_MAPPINGS: AliasMap = {
-  id: ['ID'],
+  id: ['ID', 'ID (StateAbv-number. eg., NY-1)'],
```
veekas, line 8: I figured it was still useful to leave this guidance in the title, so I added it as a mapping

test/fixtures/v1-co-81657-state-utility-lowincome.json
```diff
@@ -385,7 +385,8 @@
       "program": "Rebates & Incentives",
       "program_url": "https://www.walkingmountains.org/sustainability-hub/energy-efficiency-programs/rebates-incentives/",
       "item": {
-        "type": "smart_thermostat"
+        "type": "smart_thermostat",
+        "name": "Smart Thermostat"
```
veekas, line 389: I'm not sure why Colorado got updated as part of this PR. Happy to ditch changes to this file if I made some mistake here.

datu925: Will come by for a full review later today, but wanted to say that this is unrelated and will be fixed with https://github.com/rewiringamerica/api.rewiringamerica.org/pull/320. You can revert this file.

data/DC/incentives.json
datu925, line 1: Quite a few of these are for Other item types, and appliances that I think we've said we aren't going to serve in the API. For instance, I don't think we want to show refrigerators, since they aren't part of the electrification conversation.

As of https://github.com/rewiringamerica/api.rewiringamerica.org/pull/314, you can add a column to the spreadsheet called `Omit from API?` and any row marked TRUE will be ignored during spreadsheet-to-json processing. You can see the [AZ spreadsheet](https://docs.google.com/spreadsheets/d/11QjhvsvT2IOxNkyFYW-TpId0E5SYCixVq98jxsCRYQk/edit#gid=995688950) for an example.

data/low_income_thresholds.json
```diff
@@ -154,6 +154,48 @@
...
+        "3": 109500,
+        "4": 121700,
+        "5": 133850,
+        "6": 146000,
+        "7": 158200,
+        "8": 170350
+      }
+    },
+    "dc-dc-department-of-energy-and-environment": {
+      "/": "DOEE uses 60% AMI as the threshold for their Weatherization Assistance Program, but 80% for Solar for All. I believe that in cases like this, we opt for the higher threshold values. See https://doee.dc.gov/service/wap for 60% AMI values",
```
datu925, line 172: The right thing to do here is to define both sets of thresholds, and assign them to the appropriate records/rows. The spreadsheet-to-json script is not going to help you there, unfortunately, since it assumes that the authority is the name of the low-income threshold. You'll have to do that manually after the script runs (and unfortunately that will also prevent the sync test from being turned on for DC).

This is a known issue/gap with our low-income approahch in general, but I'm planning on addressing it next week (https://app.asana.com/0/0/1206728986903186/f).

scripts/README.md
datu925: Hmm, I've never used `npx` to run scripts. The previous version is what I use. Can you make this additive (i.e. an alternative) rather than a replacement?

datu925 (changes requested):
I'll do a second pass of the JSON itself once it's clearer which incentives are actually going to stick around.

scripts/README.md
veekas: if the previous command works for you, then I must be missing some compile step locally, because there is no `build` directory and no `.js` files in my local repo 🤔 I looked through the various scripts in `package.json` and none appear to create a `build` directory (though I'm now noticing that `yarn start` doesn't work for the same reason.

datu925: Ah – the implicit assumption behind those commands is that you've compiled first. `yarn tsc` will create that `build/` directory.

I had assumed everyone was compiling one way or another (either with yarn tsc or having configured it to run automatically in one's code editor), and I guess the `yarn start` made the same assumption.

If you can add a note indicating that you should compile first for the `node` version, then I guess that would make the advice cover all approaches.

veekas: ah, of course that's it! I searched through the code for `tsc` but didn't see it anywhere, so I had assumed that it wasn't expected. 🙃  thanks Dan

data/authorities.json
```diff
@@ -183,6 +183,21 @@
...
+        "city": "Washington, DC",
+        "county": "District of Columbia"
+      }
+    },
+    "state": {
+      "dc-dc-department-of-energy-and-environment": {
+        "name": "DC Department of Energy and Environment"
+      }
+    },
+    "utility": {}
```
oyamauchi, line 199: This will make it so that there are no utility options available in the selector, which I think will be confusing. (The options presented to users are the intersection of this set and the results from `zip_to_utility` in sqlite.) I know there are no incentives offered by the utility in DC, but we still ought to fix this.

The results from sqlite:
```
sqlite> select * from utilities where utility_id like 'dc-%';
utility_id                 name
-------------------------  ----------------------
dc-pepco                   Pepco
dc-washington-gas          Washington Gas
dc-potomac-electric-power  Potomac Electric Power
```

The only one you need to add here is Pepco. (Potomac Electric Power is its full legal name, but it's branded to customers as Pepco.)

You'll also need to follow the Utility Data section in `scripts/README.md` to modify the script that generates the utility data, to exclude Washington Gas and override Potomac Electric Power -> Pepco.

oyamauchi (changes requested):

data/authorities.json
```diff
@@ -183,6 +183,25 @@
       }
     }
   },
+  "DC": {
+    "city": {
+      "dc-dc-sustainable-energy-utility": {
+        "name": "DC Sustainable Energy Utility",
+        "city": "Washington, DC",
```
datu925, line 190: This value is load-bearing and needs to match the city names that appear in the zips.csv file. I think the name you want is just Washington.

That said, look at the next comment before you make this change since it might make it moot.

scripts/incentive-spreadsheet-registry.ts
```diff
@@ -44,6 +44,16 @@ export const FILES: { [ident: string]: IncentiveFile } = {
     enHeader: 'Program Description (guideline)',
     esHeader: 'Program Description (Spanish)',
   },
+  DC: {
+    filepath: 'data/DC/incentives.json',
+    sheetUrl:
+      'https://docs.google.com/spreadsheets/d/e/2PACX-1vQb8t0MXV32iNBujzP4s5saPZl7qpaBcVckDCqSwzHHKAQVdaQtBvE2FgPCv1XvwDGN5ORAlXtKMqmp/pub?gid=995688950&single=true&output=csv',
+    headerRowNumber: 2,
+    idHeader: 'ID',
```
datu925, line 52: I just merged a PR getting rid of some of these fields, so you'll need to delete them when you merge.

data/DC/incentives.json
```diff
@@ -0,0 +1,330 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 50
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$50 rebate for  ENERGY STAR® qualified smart thermostats."
```
datu925, line 90: Nit: there are two spaces in between `for` and `ENERGY`.

data/DC/incentives.json
```diff
@@ -0,0 +1,330 @@
...
+    "authority": "dc-dc-sustainable-energy-utility",
+    "payment_methods": [
+      "assistance_program"
+    ],
+    "item": "heat_pump_air_conditioner_heater",
+    "program": "dc_dCSustainableEnergyUtility_affordableHomeElectrificationProgram",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 10200
+    },
```
datu925, line 155: This incentive is a bit wacky. It makes more sense to express as a percentage with 100% of cost with a $10,200 maximum than as it's framed now.

data/DC/incentives.json
```diff
@@ -0,0 +1,330 @@
...
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "The affordable home electrification program can help you make the switch from natural gas and electrify your home heating and water heating systems."
```
datu925, line 161: Descriptions should include some mention of the financial terms involved. I'd rewrite this as something like:

100% of costs covered up to $10,200 for switching to electric heating and hot water systems. Income-qualified residents only.

data/DC/incentives.json
```diff
@@ -0,0 +1,330 @@
...
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "$375-$700 rebate for qualified variable speed (ductless) mini-split heat pumps systems."
+    },
+    "start_date": 2023,
+    "end_date": 2024
+  },
+  {
+    "id": "DC-12",
```
datu925, line 144: Incentives DC-12 through DC-20 are odd as currently represented, but I'm not sure what to do about it. It could just be that they changed the program from 2023 to 2024 so the spreadsheet isn't reflecting reality any longer – but if so, the end date should be 2023.

1. I don't see where the any of the numbers/maximums came from.
2. The descriptions should mention that it's for income-qualified residents only, and refer to the financial terms. See comment below for an example of how I'd rewrite it. None of the incentives are tagged with low-income though, which is also an issue.
3. If we're going to split the rows by appliance type, which the spreadsheet has done, the descriptions should mention the appliance type.

data/DC/incentives.json
```diff
@@ -0,0 +1,330 @@
...
+    "amount": {
+      "type": "percent",
+      "number": 1
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Provides low-income residents with technical and financial assistance to help reduce their energy bills by making their homes more energy efficient."
```
datu925, line 299: Also not like most of our descriptions – lacks specificity around what work is done and what the financial arrangement is.

data/DC/incentives.json
```diff
@@ -0,0 +1,330 @@
...
+    "amount": {
+      "type": "percent",
+      "number": 1
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Local solar contractors design and install solar photovoltaic (PV) systems at no cost to income-qualified homeowners."
```
datu925, line 324: This largely has the right info but should be framed for the customer perspective: e.g. 100% of costs covered for income-qualified homeowners to have solar photovoltaic (PV) systems installed by local contractors.

datu925 (approved):

datu925:
Approved to unblock beta testing today. The most pressing change to get right in this PR is the City rename in authorities.json.

Please address my other comments in a follow-up PR.

oyamauchi (approved):

PR #314, opened 2024-02-28
Add omit_to_api to our schema to filter out unwanted records.

datu925:
I figured we'd need this eventually, but it seems like now is the time since:
1. PEP folks requested this so that when there are records in the spreadsheet that shouldn't appear in the JSON, they don't have to keep deleting the generated output from the spreadsheet-to-json flow.
2. Upcoming test/script will verify equivalence of spreadsheet and JSON data, so we do need a way to mark that some records can be skipped.

Design choices:
- it's omit_from_api rather than serve_in_api because serving should be the default
- we drop these records entirely now, though as a follow-up, I'll probably have them returned as a separate category so they can be properly accounted for.

### DIFF
--- a/scripts/lib/format-converter.ts
+++ b/scripts/lib/format-converter.ts
@@ -10,12 +10,16 @@ const ajv = new Ajv({ allErrors: true, coerceTypes: 'array' });
 // TODO: consolidate with spreadsheet-standardizer.ts constant
 // and generate both directly from CollectedIncentive schema.
 const ARRAY_FIELDS = ['data_urls', 'payment_methods', 'owner_status'];
+const BOOL_FIELDS = ['omit_from_api'];
 
 const validate = ajv
   .addSchema(LOCALIZABLE_STRING_SCHEMA)
   .compile(COLLECTED_DATA_SCHEMA);
 
 type NestedKeyVal = { [index: string]: NestedKeyVal };
+export type CollectedIncentivesWithErrors = Partial<CollectedIncentive> & {
+  errors: object[];
+};
 
 /**
  * Call this function to perform validation on the collected records.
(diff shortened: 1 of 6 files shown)


scripts/lib/format-converter.ts
```diff
@@ -101,3 +112,9 @@ export function flatToNested(
   }
   return objs;
 }
+
+function coerceToBoolean(input: string): boolean | string {
+  if (input.toLowerCase() === 'true') return true;
+  if (input.toLowerCase() === 'false') return false;
+  return input;
```
mmezher, line 119: do we ever expect this to return the string in a non-error way? If not, then could we make this an error instead?

mmezher: (and if so, should there be a test for it?)

mmezher (changes requested):
Some question about the coerceToBoolean function.

scripts/lib/format-converter.ts
datu925: The output of this feeds into more robust schema validation with `ajv`, which will fail if the value is a non-convertible boolean.

The goofy thing is that ajv can handle `"true"` (the string) to `true` (the boolean) in its conversion, but it can't handle `"TRUE"`, which is how the csv/google sheets booleans are represented. So I needed some conversion method to basically help ajv along. My first version of this wasn't even this function, it was literally just lowercasing the string (which worked), but I decided that might be more confusing.

It seemed a bit redundant to have our own error handling here, hence the "minimal effort" part of the conversion.

Given that extra context, do you still think we should handle it too or is it okay as is? I can definitely add some of the above context in comments as an alternative.

mmezher: Adding context in comments seems reasonable - I think it's a little close to a code smell for this function to exist with only further validation downstream of the non-convertible boolean, but tradeoff might be fine for this case.

mmezher (approved):
Please add comment about function use, otherwise LGTM

PR #315, opened 2024-02-28
Add Github Actions for freshness check

ayangster:


scripts/check-data-freshness.ts
```diff
@@ -0,0 +1,97 @@
...
+    // If a non-2xx response status exists, return it.
+    if (error.response) {
+      /**
+      console.log(
+        source,
+        ' has an non-OK status code of: ',
+        error.response.status,
+        'with URL',
+        link,
+      );
```
mmezher, line 28: Could you remove this commented-out code? I wasn't sure the best place to add the responses, but I think tap's comments are good enough. (here and below)

mmezher (approved):
LGTM - if you can remove the commented out lines in the availability function, would be great!

PR #317, opened 2024-02-28
Add sync test that confirms that spreadsheet and JSON data is aligned.

datu925:
This is more-or-less what I demo'd on Monday. We're pretty close to a solution of this shape here, so putting out this PR to spur discussion on some of the finer points – or to surface major philosophical objections. The main files to review are the test itself and the readme; the other changes are a bit ancillary.

This test – which replaces a weaker version that verified all data matched our collected schema – validates that all rows in a spreadsheet are in sync with checked-in JSON. It does so by basically running the spreadsheet-to-json flow and verifying that the output is the same as what's checked in. That means we will start checking in any records that fail our schema validations, though I think that's a good thing as it's a way of documenting gaps in our process.

This can't be a blocking test since it can start to fail due to external factors (namely someone updating the spreadsheet without changing the JSON!), and we don't want to halt other PRs because of that. My gut reaction is that we can start with it being run as needed when updating a state. Eventually, we can add an automatic daily (or more frequent) run – this won't stop anyone's workflow, but it'll give the team a heads-up that something needs to be attended to.

We can turn this on gradually, state by state, only for the states where maintaining the test is easy. I suspect that only CO passes right now, but CT and AZ are close.


test/data/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. fail CollectedIncentive validation. This means it doesn't even meet our schema for collected data in spreadsheet form. These go in the `incentives_invalid_collected.json` file for that state.
+2. fail StateIncentive validation. These records are valid spreadsheet records but something is failing in the process of converting them to servable JSON. These go in the `incentives_invalid_state.json` file for that state.
+3. be valid, servable JSON. These go in the `incentives.json` file for that state.
+
+Eventually, we'd want all spreadsheet records go in category #3, but we're not there yet. In the meantime, categories #1 and #2 are for meeting the spreadsheets as they are and making sure they don't get worse (worse meaning more out of sync from our schemas and data).
+
+### Registering a state and fixing failures the first time
+
+To add a new state, flip the `runSpreadsheetHealthCheck` boolean in `scripts/incentive-spreadsheet-registry.ts` to true. Then run the test. It'll almost certainly fail.
```
mmezher, line 19: Not all states currently have this boolean set - should this read "set the `runSpreadsheetHealthCheck` boolean..." instead, to capture this distinction?

test/data/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Assuming it does, run the `incentive-spreadsheet-to-json.ts` script (documented in `scripts/README.md`) to overwrite the existing JSON files, and then look at the diffs to see what needs to be changed. A diff means the spreadsheet and checked-in data are out of sync, and since the checked-in data is automated downstream processing of the spreadsheet, it's likely the spreadsheet that needs to be updated.
```
mmezher, line 21: This is saying to look at the diff between the original JSON file(s) and the generated ones, correct? 

I think that something to think of here is that it is possible someone has updated/added rows to the spreadsheet, and that could be why this is failing as well. I'm not sure if it is worth calling out that the spreadsheet is likely to need updating vs referring to the "potential reasons" section below.

test/data/README.md
```diff
@@ -0,0 +1,43 @@
...
+- Wordsmith-ing or minor changes might have been made to the JSON and never propagated back to the spreadsheets. This is understandable, but we don't want this in the future.
+- Authorities or programs are commonly renamed, because our rules for converting authority or program names into string identifiers has changed over time. You might need to go back one step easlier and re-run the _entire_ spreadsheet-to-json flow, starting with `generate-misc-state-data.ts` (again, see the `scripts/README.md`). In particular, we now actually require utility Authorities to appear in the utilities database, meaning that some of the Authority Names in the spreadsheet are wrong.
+
+There are some known pain points where our automation isn't good enough. Low-income thresholds don't work well now when the threshold is the state's default. You'll probably need to manually correct some records each time you re-run the flow. This is a known issue.
```
mmezher, line 33: do we have an Asana ticket to track this?

test/data/README.md
datu925: Yes, set is more accurate terminology than fix. Will correct.

test/data/README.md
datu925: I just filed https://app.asana.com/0/0/1206728986903186/f for this. But I'm planning to work on it sooner rather than later. Owen and I chatted about a few different options but I'm still weighing pros and cons.

test/data/README.md
datu925: Ack. I pretty much agree with all of this and will make some changes to reflect that.

test/data/spreadsheets-health.test.ts
```diff
@@ -23,22 +28,55 @@ test(
...
+          invalidCollectedPath,
+        )
+          ? JSON.parse(fs.readFileSync(invalidCollectedPath, 'utf-8'))
+          : [];
+        const previousStateExceptions: string = fs.existsSync(invalidStatePath)
+          ? JSON.parse(fs.readFileSync(invalidStatePath, 'utf-8'))
+          : [];
+
+        // We expect equality of any invalid records, since humans shouldn't be
+        // editing these anyway.
```
mmezher, line 49: Isn't it possible that new, invalid records have been added to the spreadsheet? I agree that humans shouldn't be editing the invalid records.

mmezher (commented):
Some documentation questions.

test/data/spreadsheets-health.test.ts
datu925: If so, then those records should be added to the golden checked-in file.

The comment is supposed to contrast from the approach further below where we only test equality of some fields, but let me know if it's misleading.

mmezher:
Resolve conflicts? Happy to take another look.

scripts/README.md
```diff
@@ -97,3 +97,45 @@ When adding support for a new state, you should vet and clean up the utility dat
...
+
+Sometimes, you might find records that are in the spreadsheets but not in the JSON because they're really far away from being convertable – perhaps they're an Item type that we don't know how to capture, or something that we're far away from being able to model. In that case, you can leave it in those files; just check them in. They're now clearly documented gaps in our data processing pipeline.
+
+There are a lot of potential reasons for diffs in the valid JSON:
+
+- someone may have added or subtracted rows in the spreadsheet, in which case you should expect whole records to be added or removed from the JSON. It could also be that those records were always the spreadsheet, but when the JSON conversion process was manual, someone left them out intentionally for editorial reasons. You can use the `Omit from API?` column in the spreadsheet to ignore records if you don't want them to be served by the API.
```
mmezher, line 127: nit: always in the

datu925: Fixed.

scripts/README.md
```diff
@@ -97,3 +97,45 @@ When adding support for a new state, you should vet and clean up the utility dat
...
+
+Assuming it does, the easiest way to resolve any diffs is to run the `incentive-spreadsheet-to-json.ts` script (documented in `scripts/README.md`) to overwrite the existing JSON files, and then look at the diffs to see what needs to be changed. A diff means the spreadsheet and checked-in data are out of sync – see potential reasons for diffs below.
+
+Our goal is that converting collected spreadsheet data into JSON is automatable, so if something is failing, it is a problem we eventually need to solve. If there are records in categories #1 or #2 above, spend some time trying to fix the issues by looking at the `errors` field in those files. Updating the spreadsheet or occasionally the codebase to fix errors is preferable to checking in the invalid record.
+
+Sometimes, you might find records that are in the spreadsheets but not in the JSON because they're really far away from being convertable – perhaps they're an Item type that we don't know how to capture, or something that we're far away from being able to model. In that case, you can leave it in those files; just check them in. They're now clearly documented gaps in our data processing pipeline.
```
mmezher, line 123: Can you be more explicit here about what files are being referred to here? Something like "In that case, you can leave the generated invalid records in the state's incentives_invalid_state.json".

mmezher (approved):
Other than documentation change, LGTM. Think other comments I had were outdated or misinformed.

PR #322, opened 2024-02-29
[RFC] Add "utilities" list field to incentives

oyamauchi:
I've been messing around with this for days, and it's time to just get
something out there and see what people think. This just does the
modification to the PRPA incentives in Colorado, since it was the
simplest and I'm not fully convinced by this approach.

I find this awkward in a few ways:

- I can't picture how the mapping from spreadsheet to JSON would work
  here. For spreadsheet purposes it seems most straightforward just to
  have "PRPA", but where in the spreadsheet-to-JSON process would this
  new field be introduced?

- It's duplicative in data; every incentive from one of these
  consortiums will have the same list of utilities. Not a huge
  problem, but still.

- It's duplicative in functionality; there'd now be two ways to
  specify an incentive that's offered to customers of a single
  utility.

- It doesn't cleanly extend to groups of counties (a thing that exists
  in NV, for example). Sure you could add an analogous `counties`
  field, but what would it mean for an incentive to have both a
  `counties` and a `utilities` field?

### Alternatives?

Alternative approaches to this all fall into the category of "define
authority groups somehow". That solves the duplicative-data problem,
but introduces all sorts of awkwardness in deciding how to represent
the groups.

- Is a group of authorities itself a kind of authority, or a distinct
  concept?

- Could a group include different kinds of authority, like some
  utilities and some counties? What would that even mean -- "one of
  these utilities AND one of these counties", or "one of these
  utilities OR one of these counties"?

I'm starting to form a thought on this that what's making things
difficult is the overlap of namespaces: utilities, cities, and
counties are each a namespace of their own, and an authority is its
own identity (name, logo) that can consist of one or more of the
above. That's just an awkward thing to represent.

### DIFF
--- a/data/CO/incentives.json
+++ b/data/CO/incentives.json
@@ -1808,8 +1808,14 @@
   },
   {
     "id": "CO-81",
-    "authority_type": "utility",
+    "authority_type": "state",
     "authority": "co-platte-river-power-authority",
+    "utilities": [
+      "co-estes-park-power-and-communications",
+      "co-fort-collins-utilities",
+      "co-longmont-power-and-communications",
+      "co-loveland-water-and-power"
+    ],
     "payment_methods": [
       "rebate"
     ],
@@ -1830,8 +1836,14 @@
   },
   {
     "id": "CO-82",
-    "authority_type": "utility",
+    "authority_type": "state",
     "authority": "co-platte-river-power-authority",
+    "utilities": [
+      "co-estes-park-power-and-communications",
+      "co-fort-collins-utilities",
+      "co-longmont-power-and-communications",
+      "co-loveland-water-and-power"
+    ],
     "payment_methods": [
       "rebate"
     ],
@@ -1852,8 +1864,14 @@
   },
   {
     "id": "CO-83",
-    "authority_type": "utility",
+    "authority_type": "state",
     "authority": "co-platte-river-power-authority",
+    "utilities": [
+      "co-estes-park-power-and-communications",
+      "co-fort-collins-utilities",
+      "co-longmont-power-and-communications",
+      "co-loveland-water-and-power"
+    ],
     "payment_methods": [
       "rebate"
     ],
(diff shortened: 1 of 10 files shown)


datu925:
I like the PR as a way of spurring conversation, though this does feel complex enough that maybe it's worth having a doc to fully flesh out the tradeoffs?

My initial thoughts are:
- this is another instance of authority being only loosely connected to eligibility
- that nuance is also something third-party API clients may want to know. We may want to return it in the API rather than just rely on it for our own eligibility processing
- as you've pointed out, this isn't unique to utilities

My starting bid is that we should generalize the approach to a more general optional "eligibility" concept that is distinct from authority, though in the standard case when that column isn't populated, it's inferred from the authority type. Eligibility would be filled out in free-text form in the spreadsheet by data collectors. When it's time to produce JSON, we use a similar treatment to low-income thresholds or incentive relationships – it will need to be configured in code somewhere. It might include `utilities` or `counties` or `cities` or perhaps combinations (though what that means is a good question as you've noted).

There's a lot more to unpack but that's my gut instinct.

rpartridge:
Overall this makes sense but I agree it could get messy when we throw counties into the mix. From my understanding, we use authorities to determine incentive eligibility, right? If so I wonder if we could be more explicit about that in the authority definitions, rather than needing to list out all of the eligible utilities or counties for every incentive item. 

Something like:
```json
// In data/authorities.json
"CO": {
  "state": {},
  "utility": {},
  // Not sure about naming
  "group": {
    "co-platte-river-power-authority": {
      "name": "Platte River Power Authority",
      "eligibility": {
        "utilities": ["co-estes-park-power-and-communications", "co-fort-collins-utilities"],
        "counties": ["White Pine"]
      }
    }
  }
},
```

Then when a user fetches incentives we could use the eligibility rules to determine if they qualify. I'm not sure how hard that would be and how well it translates to the spreadsheet to JSON conversion though. And take this with a grain of salt since I'm still new to the codebase!

oyamauchi:
OK, I started to write up a proposal doc but in the process I think I found an approach that I like more than the others. I'll update this PR to show it. I think seeing @rpartridge 's example code is what unstuck me, so thanks!

oyamauchi:
Actually, closing in favor of #329

PR #323, opened 2024-02-29
Add "zip" and "address" top-level params

oyamauchi:
## Description

The `location` param being a key-value map makes it a little awkward
for some HTTP tooling. See https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1707244817589029

Instead, put the `zip` and `address` params at the top level. This PR
adds them and keeps `location` as-is, until we can switch over the
frontend, at which time we'll remove `location`. One of the three
params is required.

## Test Plan

`yarn test`. Tested the `address` param manually with:

`curl 'http://localhost:3000/api/v1/calculator?owner_status=homeowner&household_income=50000&tax_filing=single&household_size=1&address=82%20Smith%20St,Providence'`

and made sure RI-specific incentives show up.

### DIFF
--- a/src/routes/v1.ts
+++ b/src/routes/v1.ts
@@ -46,7 +46,7 @@ export default async function (
   fastify: FastifyInstance & { sqlite: Database },
 ) {
   async function fetchAMIsForLocation(
-    location: APIRequestLocation,
+    location: Partial<APIRequestLocation>,
   ): Promise<IncomeInfo | null> {
     if (location.address) {
       // TODO: make sure bad addresses are handled here, and don't return anything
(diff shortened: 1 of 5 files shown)


src/routes/v1.ts
```diff
@@ -77,12 +77,16 @@ export default async function (
     { schema: API_CALCULATOR_SCHEMA },
     async (request, reply) => {
       const language = request.query.language ?? 'en';
-      const incomeInfo = await fetchAMIsForLocation(request.query.location);
+      const queryLocation = request.query.location ?? {
```
veekas, line 80: good thought to continue to accept the old param to decouple the work

veekas (approved):

PR #327, opened 2024-03-04
Make pass exception for 403 status code

mmezher:
Currently, Cloudflare is being used by efficiency Vermont to prevent attacks on their website. Unfortunately, this means regular scraping techniques are denied access. We'll need to do some exploration of how to work around this safely, but for now, these should likely not be considered failing.


scripts/check-data-freshness.ts
```diff
@@ -15,6 +15,7 @@ async function checkUrlDataAvailability(
     method: 'get',
     url: link,
     timeout: 10000,
+    validateStatus: () => true,
```
ayangster, line 18: What's the purpose of this argument?

ayangster:
Curious about the argument but otherwise LGTM!

ayangster (approved):

scripts/check-data-freshness.ts
mmezher: This allows all non-null, non-2xx statuses to count as being “okay” and therefore have its status returned. What I found was that currently, non-2xx weren’t getting treated as errors with statuses, but returning no status. This allows us to do some handling of returned, non-2xx status codes ourselves.

PR #328, opened 2024-03-04
Add GA incentives

rpartridge:
Adds Georgia incentives: https://docs.google.com/spreadsheets/d/1DH5Q6IzH-UhevG0jiQb7IrchU7GGkVsPoVMRTmS7tmU/edit#gid=995688950

I did end up doing a data refresh for these incentives since there were only 7, so the data should be up to date


data/low_income_thresholds.json
```diff
@@ -154,6 +154,21 @@
       }
     }
   },
+  "GA": {
```
rpartridge, line 157: GA doesn't have any income-qualified programs but the schema test requires a default is filled out. This just uses the federal poverty guidelines

data/GA/incentives.json
```diff
@@ -0,0 +1,170 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 250,
+      "maximum": 250
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Rebate for 50% of cost up to $250 for attic insulation. Addition of R-38 or greater required."
```
datu925, line 20: This should be type percentage with number 0.5, but the 250 maximum is correct.

Same true for all of the above (these are errors in the original spreadsheet).

data/GA/incentives.json
```diff
@@ -0,0 +1,170 @@
...
+    },
+    "start_date": 2023,
+    "end_date": 2025,
+    "bonus_available": true
```
datu925, line 24: The data collector put some information in the "Bonus Details" column, but it all looks like it's related to how the incentives should be combined, which should go under Stacking Details instead.

1. Move the data into Stacking Details. Once Bonus Details is blank, `bonus_available` boolean will stop appearing when you rerun the script.
2. It looks like there is a maximum across several incentives, and we do have a way to represent that, but you'll need to add another file. See data/CO/incentive_relationships.json for an example of how to configure a maximum.

datu925 (changes requested):
Thanks for doing the refresh. This needs a few changes mostly due to spreadsheet issues, but otherwise looks good.

data/GA/incentive_relationships.json
datu925, line 1: Sorry, you also have to link them into (and configure as appropriate):
src/data/state_incentive_relationships.ts
test/data/schemas.test.ts

datu925 (changes requested):
Almost there!

data/GA/incentive_relationships.json
rpartridge: ahh right! fixed, thanks

datu925 (approved):

PR #329, opened 2024-03-04
[RFC] Add concept of "geo groups"

oyamauchi:
This is where I got to following the comments on #322.

- There's a new concept that I'm calling "geo groups" (open to
  bikeshedding). Each one can contain a set of cities, a set of
  counties, and a set of utilities (in the form of their IDs from
  authorities.json).

  Geo groups are defined in a new file, `data/geo_groups.json`. I
  considered defining them in authorities.json, but I think it's
  better to clearly separate the concept of authority from geographic
  eligibility; that was part of the whole point of this, and I kind of
  lost sight of it before. A geo group is not an authority (although
  in practice there's probably very significant overlap); it's just a
  set of utilities/cities/counties.

  One potentially confusing thing is that geo group IDs are a totally
  separate namespace from all the authority namespaces. The fact that
  the authority ID and geo group ID are both `co-platte-river-power-authority`
  here is incidental. In fact, maybe they should be intentionally made
  different somehow, to emphasize that?

- There's a new authority type I'm calling `other`. It's meant for
  types of authority that don't have an inherent geographic
  restriction, which all the others do. Incentives with this authority
  type must have a `geo_group` key that references a geo group.

- When an incentive is of type `other`, its geo group is checked to
  make sure that the request params match each component. The
  request's `utility` must match one of the geo group's `utilities`
  items, etc.

  That means the combining semantic, for a geo group that defines more
  than one set of authorities, is "AND". If you need "OR" semantics
  (e.g. one of these utilities OR one of these counties), create two
  geo groups (one with only utilities and one with only counties),
  duplicate the incentive (one referring to each of the two geo
  groups), and add a mutual exclusion between the two copies of the
  incentive. (I have no idea if this will ever come up, but it's
  handled by this scheme.)

- The common cases (a utility offers an incentive to its customers, a
  city or county offers an incentive to its residents) are unchanged.

I think this will fit cleanly into the existing spreadsheet-to-JSON
process of defining authorities, though I'm not 100% sure. Certainly
it's more likely to do so than my first attempt was.

### DIFF
--- a/data/geo_groups.json
+++ b/data/geo_groups.json
@@ -1,6 +1,6 @@
 {
   "CO": {
-    "co-platte-river-power-authority": {
+    "co-group-platte-river-power-authority": {
       "utilities": [
         "co-estes-park-power-and-communications",
         "co-fort-collins-utilities",
--- a/data/geo_groups.json
+++ b/data/geo_groups.json
@@ -1,0 +1,12 @@
+{
+  "CO": {
+    "co-platte-river-power-authority": {
+      "utilities": [
+        "co-estes-park-power-and-communications",
+        "co-fort-collins-utilities",
+        "co-longmont-power-and-communications",
+        "co-loveland-water-and-power"
+      ]
+    }
+  }
+}
--- a/scripts/update-fixtures.sh
+++ b/scripts/update-fixtures.sh
@@ -206,6 +206,7 @@ curl \
 &items=heat_pump_water_heater\
 &utility=co-xcel-energy" \
   | jq . > test/fixtures/v1-80517-xcel.json
+
 curl \
   "http://localhost:3000/api/v1/calculator\
 ?location\[zip\]=80517\
--- a/scripts/update-fixtures.sh
+++ b/scripts/update-fixtures.sh
@@ -178,6 +178,7 @@ curl \
 &household_size=4" \
   | jq . > test/fixtures/v1-15289-homeowner-80000-joint-4.json
 
+
 # TODO: Remove beta states argument when DC is fully launched.
 curl \
   "http://localhost:3000/api/v1/calculator\
@@ -190,3 +191,32 @@ curl \
 &authority_types=state\
 &authority_types=city" \
   | jq . > test/fixtures/v1-dc-20303-state-city-lowincome.json
+
+curl \
+  "http://localhost:3000/api/v1/calculator\
+?location\[zip\]=80517\
+&include_beta_states=true\
+&owner_status=homeowner\
+&household_income=80000\
+&tax_filing=single\
+&household_size=1\
+&authority_types=state\
+&authority_types=utility\
+&authority_types=other\
+&items=heat_pump_water_heater\
+&utility=co-xcel-energy" \
+  | jq . > test/fixtures/v1-80517-xcel.json
+curl \
+  "http://localhost:3000/api/v1/calculator\
+?location\[zip\]=80517\
+&include_beta_states=true\
+&owner_status=homeowner\
+&household_income=80000\
+&tax_filing=single\
+&household_size=1\
+&authority_types=state\
+&authority_types=utility\
+&authority_types=other\
+&items=heat_pump_water_heater\
+&utility=co-estes-park-power-and-communications" \
+  | jq . > test/fixtures/v1-80517-estes-park.json
--- a/src/data/geo_groups.ts
+++ b/src/data/geo_groups.ts
@@ -1,0 +1,40 @@
+import fs from 'fs';
+import { FromSchema } from 'json-schema-to-ts';
+import { STATES_PLUS_DC } from './types/states';
+
+const GEO_GROUP_SCHEMA = {
+  type: 'object',
+  properties: {
+    utilities: { type: 'array', items: { type: 'string' }, minItems: 1 },
+    cities: { type: 'array', items: { type: 'string' }, minItems: 1 },
+    counties: { type: 'array', items: { type: 'string' }, minItems: 1 },
+  },
+  additionalProperties: false,
+  anyOf: [
+    { required: ['utilities'] },
+    { required: ['cities'] },
+    { required: ['counties'] },
+  ],
+} as const;
+
+export type GeoGroup = FromSchema<typeof GEO_GROUP_SCHEMA>;
+
+export const GEO_GROUPS_SCHEMA = {
+  type: 'object',
+  propertyNames: {
+    type: 'string',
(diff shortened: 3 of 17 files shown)


test/data/schemas.test.ts
```diff
@@ -187,6 +192,21 @@ test('state incentives JSON files match schemas', async tap => {
...
         );
       }
+      if (incentive.authority_type === AuthorityType.Other) {
+        tap.hasProp(
+          incentive,
+          'geo_group',
+          `authority_type 'other' must include a geo group (id ${incentive.id})`,
+        );
+
+        const stateGroups = GEO_GROUPS_BY_STATE[stateId];
```
datu925, line 202: Performance isn't a concern, but I'd move this up to where we define `authorities` since it's the same for every record in each state.

test/data/schemas.test.ts
```diff
@@ -187,6 +192,21 @@ test('state incentives JSON files match schemas', async tap => {
...
+        tap.equal(incentive.authority_type, AuthorityType.Other);
```
datu925, line 203: Don't we know this is true based on being in this conditional branch?

src/data/state_incentives.ts
datu925, line 1: I suspect that like low-income, this is a case where we want to gather free-text info about the groups in the spreadsheets in the initial collection stage and then codify it later. If so, you should add a field to the CollectedIncentive schema and aliases in scripts/lib/spreadsheet-mappings.ts, though fine to do in a follow-up if you prefer to keep it separate.

src/data/state_incentives.ts
```diff
@@ -126,6 +126,7 @@ export type DerivedFields = {
   agi_max_limit?: number;
   agi_min_limit?: number;
   authority: string;
+  geo_group?: string;
```
datu925, line 129: Re: name bikeshedding, I don't love geo_group because if you don't know what it is already, you won't guess what it is from the name.

In my mind this relates to eligibility, so I like eligibility_group. Or if we want to emphasize that it is a particular subset of eligibility, maybe geo_eligibility_group.

test/data/schemas.test.ts
```diff
@@ -187,6 +192,21 @@ test('state incentives JSON files match schemas', async tap => {
...
+        tap.hasProp(
+          stateGroups,
+          incentive.geo_group!,
+          `nonexistent geo_group (${stateId}, id ${incentive.id})`,
+        );
+      }
```
datu925, line 209: Do we want to ensure that *only* Other authorities define geo_groups? Or is it intentional so that we can separate authority and eligibility even for non-Other authority types?

datu925 (changes requested):
This is a good evolution and I have some small things but overall 👍.

Re: spreadsheet-to-json flow, any process works well where:
1. the additional group metadata is configured once by a human in a separate file, and 
2. the key to that metadata can be derived from info in the incentive so that the script can pull in the right metadata

Both are satisfied here (assuming we always make the key out of some derivative of the authority name) so I don't see any problems. Fine to add `group_` or similar prefix to the group name if you want to separate the namespace from other authorities.

data/geo_groups.json
```diff
@@ -0,0 +1,12 @@
+{
+  "CO": {
+    "co-platte-river-power-authority": {
+      "utilities": [
+        "co-estes-park-power-and-communications",
+        "co-fort-collins-utilities",
+        "co-longmont-power-and-communications",
+        "co-loveland-water-and-power"
+      ]
```
rpartridge, line 9: A pre-requisite for adding utilities/cities/counties here would be that they're already defined as authorities on their own, right?

src/data/state_incentives.ts
rpartridge: I'm a fan of `geo_eligibility_group`

rpartridge (commented):
Approach makes sense to me!

Do we already have mappings for zip code > city and zip code > county that we can leverage? Doesn't apply to this utilities example in Colorado, just wondering for the future

scripts/update-fixtures.sh
```diff
@@ -189,4 +190,33 @@ curl \
...
+&household_income=80000\
+&tax_filing=single\
+&household_size=1\
+&authority_types=state\
+&authority_types=utility\
+&authority_types=other\
+&items=heat_pump_water_heater\
+&utility=co-xcel-energy" \
+  | jq . > test/fixtures/v1-80517-xcel.json
+curl \
```
datu925, line 209: nit: add a line in between for better readability.

src/lib/incentives-calculation.ts
```diff
@@ -381,13 +381,9 @@ export default function calculateIncentives(
   const authorities: AuthoritiesById = {};
   if (stateAuthorities) {
     incentives.forEach(i => {
-      if (
-        'authority' in i &&
-        i.authority &&
-        (i.authority_type === 'state' || i.authority_type === 'utility')
-      ) {
+      if ('authority' in i && i.authority && i.authority_type !== 'federal') {
```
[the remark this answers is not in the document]
datu925: Thoughts?

src/data/geo_groups.ts
datu925, line 1: Optional: rename both the file and the various constants here to be geo_eligibility_groups instead of geo_groups, since that's what the field is called now.

datu925 (approved):
Some small things still, but nothing worth blocking over.

src/lib/incentives-calculation.ts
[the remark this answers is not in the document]
oyamauchi: (Oh ugh apparently I didn't actually post this before)

Unfortunately `authority` isn't necessarily defined because `i` is a `CalculatedIncentive`, which is a supertype of `IRAIncentive`, which doesn't have the field. And if you don't have the `in` check, TS complains that `authority` doesn't exist on `CalculatedIncentive`.

I guess it would be cleaner if all the IRA incentives had `authority: null`, then this could just be `if (i.authority)`, skipping the "is federal" check. But I'd rather do that separately.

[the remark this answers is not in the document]
datu925: Ah gotcha, I'd forgotten about IRAIncentives. I don't feel strongly about the cleanup – this just seemed like low-effort refactoring but ack that that's not the case.

src/data/geo_groups.ts
oyamauchi: Actually, what if we leave the file/concept as "geo group" and name the field `eligibility_geo_group`? That keeps the "group" concept uncoupled from eligibility. (I doubt there'd be any other application for this concept, but you never know.)

PR #330, opened 2024-03-04
Make programs updater repeatable.

datu925:
This is part of making the `generate-misc-state-data.ts` script fully automated and repeatable. There are some diffs in structure/behavior here I wanted to call out explicitly:

- I split programs.json into federal_programs.json and state_programs.json. This seemed like the least terrible way to get proper sorting of programs given that there are some IRA programs that should(?) always be separate. I didn't want to hard-code either the number of IRA incentives or the specific names since those approaches feel really brittle.

- this does change the program order in the file. Currently, we have it alphabetized by state, but within state, the programs appear in the same order as they occur in the spreadsheet. After this change, the programs will be purely alphabetical (which still sorts them by state). If we have a reason to prefer the more nuanced sorting of the current state, it's not a big deal and I'm happy to implement it. As a result of this, programs.ts and programs.json will have different sorting order in the short-term, though I'm planning to follow up with a similar PR for programs.ts so the difference will be short-lived.


datu925:
Removing this temporarily since I think there may be a different approach that gets rid of the need for this entirely.

PR #331, opened 2024-03-04
Consolidate programs.json into programs.ts.

datu925:
Per [Slack conversation](https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1709583657513919), it'll be better to have a single data format for this program information.

This isn't strictly necessary and I am open to pushback, but I think it's better from an organizational and automation perspective to also split up these files by state, which I've done here. They are getting pretty large, and being able to simply create a new file or overwrite an existing one when updating programs is easier than needing to insert in the right place.

### DIFF
--- a/data/DC/programs.ts
+++ b/data/DC/programs.ts
@@ -48,4 +48,4 @@ export const DC_PROGRAMS = {
         en: 'https://doee.dc.gov/liheap',
       },
     },
-};
+} as const;
--- a/data/DC/programs.ts
+++ b/data/DC/programs.ts
@@ -1,0 +1,51 @@
+export const DC_PROGRAMS = {
+  dc_dCSustainableEnergyUtility_dCResidentialRebates: {
+    name: {
+      en: 'DC Residential Rebates',
+    },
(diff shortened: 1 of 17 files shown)


datu925:
Also, this will require changes to the spreadsheet-to-json flow (specifically the generate-misc-state-data.ts script) but I can do those as fast follow-up once this is approved.

data/DC/programs.ts
```diff
@@ -0,0 +1,51 @@
...
+  'dc_dCDepartmentOfEnergyAndEnvironment_LowIncomeHomeEnergyAssistanceProgram(LIHEAP)':
+    {
+      name: {
+        en: ' Low Income Home Energy Assistance Program (LIHEAP)',
+      },
+      url: {
+        en: 'https://doee.dc.gov/liheap',
+      },
+    },
+};
```
oyamauchi, line 51: This needs to be `as const` (I'm surprised it works without it, actually)

oyamauchi (approved):
Seems good

PR #333, opened 2024-03-05
Adding Spanish descriptions for VT

mariabajzek:
I ran the update-descriptions script to do this, but as a result I had to revert many of the English descriptions because they are too long in the spreadsheet and don't currently match what is in the JSON. Before launching this, we should probably review the descriptions and decide whether we want to keep the longer versions (and update the UI to display them better) or keep the shorter versions and make sure the spreadsheet is back in sync.

I also had to increase the Spanish character limit in the schema test again because many of the Spanish ones are very long for VT.

Also, the spreadsheet <--> JSON translation is going to be very complicated because of the EV incentives (e.g. VT-27). They are represented as a single spreadsheet row with one description that maps to multiple JSON objects with different amounts and descriptions based on tax filing status. That will need to be fixed before this is easily maintainable going forward.

**Test plan:**
Updated the test fixture to match the new data.


data/VT/incentives.json
```diff
@@ -423,7 +441,8 @@
     "start_date": 2022,
     "end_date": 2023,
     "short_description": {
-      "en": "BED customers may receive a rebate up to $2,500 on qualifiying HPs."
+      "en": "BED customers may receive a rebate up to $2,500 on qualifiying HPs, plus an additional $500 for a second HP.",
```
datu925, line 444: There are a few rebates where `qualifying` is misspelled. Will need to be changed probably in the spreadsheet, here, and in the test fixture.

datu925 (approved):
Address comment before merging.

data/VT/incentives.json
mariabajzek: Thank you for catching that! Fixed.

PR #336, opened 2024-03-05
Use geo groups for all cases where it applies

oyamauchi:
## Description

This applies to Walking Mountains and Tri-State G&T in CO, and VPPSA
in VT.

The Walking Mountains case is not straightforward. Their eligibility
criterion isn't exactly clear to me -- it seems to be Eagle County
minus a few specific parts, and I don't think our geographic data is
sophisticated enough to capture all that -- so I've had to just say
their incentives are restricted to Eagle County.

In the process I had to add the VT utilities that are in the dataset
and are members of VPPSA. There are several more members of VPPSA (see
https://vppsa.com/members/) that aren't in the Energy Star data, so
they aren't added here. (In general I'm not quite sure what to do
about such cases -- maybe we should support _adding_ utilities to the
dataset, as well as excluding and modifying like we already do?)

## Test Plan

`yarn test`. Manual request for VT with utility =
`vt-swanton-electric` to make sure VPPSA incentives are returned.

The "exceptions" thing in the utilities test is now gone.


datu925 (approved):
Once this is merged, can you also go through and update the relevant spreadsheets so they are aligned with the JSON data?

PR #338, opened 2024-03-06
Add new update programs method now that programs.json is consolidated.

datu925:
This is admittedly a bunch of string-hacking, but I don't see a better way to do it, aside from giving up on this step being automated – which is not out of the question since it's not run as often as the actual incentive-spreadsheet-to-json script. Still, I think we might as well give it a shot.

Testing: added a bunch of unit tests, and ran on CO and AZ to ensure that we saw no diffs.


mmezher (approved):
I'm also not sure how robust this code will be long-term, but fine with checking it in since I don't see harm in having it. Might need to remove it down the road if formats change, since maintenance might be less than useful.

PR #339, opened 2024-03-06
Refactor refineCollectedData method to a separate class.

datu925:
Spreadsheet standardization and refining collected data (e.g. getting to the JSON data we serve) were in the same class for historical reasons, but these processes are pretty separate now.

#namingishard and if you have a better name than DataRefiner, happy to change it.


mmezher (approved):

mmezher (approved):
Couldn't think of a better name for this, though ideally I think it'd distinguish what a "refined" record's format is compared to the collected one (this is where is moves to the actual, served JSON format, right?)

datu925:
> Couldn't think of a better name for this, though ideally I think it'd distinguish what a "refined" record's format is compared to the collected one (this is where is moves to the actual, served JSON format, right?)

Sorry, I was too trigger-happy with the merge :)

The type we have for it is StateIncentive, which is an existing name which doesn't really call out the API-ready, servable nature of it – which is what distinguishes it from CollectedIncentive. I suppose we could define an alias for ServableIncentive but not sure it's worth it.

PR #342, opened 2024-03-07
Update scripts to handle geo_groups

oyamauchi:
## Description

- Updates generate-misc-state-data to create entries in
  geo_groups.json for any authorities of type "other", with empty
  lists to be filled in by the engineer.

  A couple of wrinkles: the standardizer that gets created in this
  script has empty value mappings, which means the check for authority
  type has to look for the spreadsheet string `Other` rather than the
  standardized `other`. I don't know why the value mappings are empty
  in this case, but looking for the unstandardized value feels wrong.

  The standardizer is also created in strict mode, which means it
  errors out because there are extra columns in the CO and VT
  spreadsheets now ("description character count"/"description word
  count" and "questions / comments" respectively). I had to turn off
  strict mode to test the changes.

- Updates spreadsheet standardizer to add `eligible_geo_group` to any
  incentive with authority_type `other`.

The intended workflow, starting from initial state translation, is:

1. The "geographic eligibility" column in the sheet is filled with
   free text describing the criterion (or it's absent).

2. Run `generate-misc-state-data`, which will generate an entry in
   geo_groups.json for every authority with an "other"-type incentive.

3. Edit geo_groups.json manually, to define what those groups are,
   based on the free text in the sheet and/or manual research.

4. If customization is needed -- such as if one "other"-type authority
   doesn't have the same geo eligibility for all their incentives --
   define the necessary geo groups manually, and update the
   spreadsheet to put the appropriate geo group ID in the "Geographic
   Eligibility" column of each incentive where it's necessary.

5. Run `spreadsheet-to-json`. It will use the value of "Geographic
   Eligibility" as the `eligible_geo_group` if it looks like a geo
   group ID, or mechanically derive it from the authority name (the
   same way as in step 2) otherwise.

## Test Plan

Go through the above workflow with CO. (Although, as noted, I had to
turn off strict mode in step 2; I also had to comment out updating the
programs files since I'm doing this without #338.)

### DIFF
--- a/test/scripts/authority-and-program-updater.test.ts
+++ b/test/scripts/authority-and-program-updater.test.ts
@@ -479,6 +479,7 @@ test('add geo groups for new state', async tap => {
         utilities: [],
         cities: [],
         counties: [],
+        notes: '',
       },
     },
     CO: {
@@ -517,6 +518,7 @@ test('add geo groups to existing state', async tap => {
         utilities: [],
         cities: [],
         counties: [],
+        notes: '',
       },
       'co-group-something': {
         utilities: ['co-some-utility'],
--- a/test/scripts/authority-and-program-updater.test.ts
+++ b/test/scripts/authority-and-program-updater.test.ts
@@ -6,7 +6,7 @@ import {
   StateToAuthorityTypeMap,
   createProgramsContent,
   maybeUpdateProgramsTsFile,
-  sortJsonAlphabeticallyByStateKey,
+  sortMapByKey,
   updateAuthorities,
 } from '../../scripts/lib/authority-and-program-updater';
 
@@ -63,7 +63,7 @@ const baseSourceFile = `import SomeClass from './some_file';
 import CT_PROGRAMS from './programs/ct_programs';
 import NY_PROGRAMS from './programs/ny_programs';
 import OtherClass from './other_file';
-  
+
 const var = 'foo';
 
 const all_programs = {
@@ -122,10 +122,7 @@ test('correctly sort state authority information by state', tap => {
       },
     },
   };
-  tap.matchOnly(
-    ordered_json,
-    sortJsonAlphabeticallyByStateKey(unorderedFixture),
-  );
+  tap.matchOnly(ordered_json, sortMapByKey(unorderedFixture));
   tap.end();
 });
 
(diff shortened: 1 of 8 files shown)


test/scripts/authority-and-program-updater.test.ts
datu925, line 1: Can you add a test(s) for the updateGroups method?

With similar methods, I've been splitting out the main logic from the file reading/writing to make it easier to test.

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -149,6 +150,19 @@ export class SpreadsheetStandardizer {
         );
       }
     }
+    if (record.authority_type === 'other') {
+      // During initial data collection, the "Geographic Eligibility" column
+      // will be filled in with free text. After misc state data is generated,
+      // which adds entries to geo_groups.json, translate the free text into
+      // structured representation in geo_groups.json, and replace the
+      // "Geographic Eligibility" free text with the ID of the geo group.
```
datu925, line 158: Ah, this is interesting. I was figuring we would just keep the free text field for the extra nuance and just use the authority column as the key all the time.

I do like this model and it works for low-income thresholds as well – in some respect even better since we can't rely on income thresholds always having an authority-based key as we've used so far.

My only thought is that maybe we should add a context or notes column to the thresholds and groups schema so that we don't have to lose any spreadsheet nuance when replacing the free text with the identifier.

datu925 (changes requested):
Main request is test coverage; otherwise looks good.

datu925:
Thanks for doing this. I fixed the strict mode issue in https://github.com/rewiringamerica/api.rewiringamerica.org/pull/338 and the programs update is merged now, so hopefully a bit less painful going forward.

Re: VALUE_MAPPINGS, feel free to add them. Up until now, the program/authority updater has only needed to standardize column names since the values in authority and program columns are free strings. This is the first case where a field value is load-bearing for the class, so makes sense to standardize those too going forward.

mmezher (changes requested):
LGTM but agree with Dan on adding test coverage and maybe adding a free-text column for backwards comparisons.

scripts/lib/spreadsheet-standardizer.ts
oyamauchi: I added an optional `notes` field to the geo group schema. It would take extra effort for generate-misc-state-data to populate it from the sheet, though, because AuthorityAndProgramUpdater doesn't keep the entire row contents. Instead it just puts in an empty string.

mmezher (approved):

datu925 (approved):

PR #345, opened 2024-03-09
Add NV geogroups and their incentives

rpartridge:
# What

Adds 4 NV geogroups now that they're supported and the incentives that each geogroup offers. NV-74 - NV-77 here: https://docs.google.com/spreadsheets/d/1w6c4lAlNGTKEuN_X00mAcK3Xs6NGM1aT5XFbhjd1CGo/edit#gid=995688950

# Why

Originally I was going to exclude these incentives because it wasn't clear from their websites what the incentives actually are, but we got an email recently from Nevada Clean Energy Fund, one of our partners, that specifically mentioned these 4 incentives. I think we need to go ahead and include them then since NCEF is going to be helping us QA and I think they'll notice if these incentives are missing. I ended up looking at the websites in more detail and the incentives seem to be that an application is required on top of meeting the income requirements, but that if an application is accepted, that resident will get all the services available free of charge. I added a note in the description about the application and marked this as `percent` with percent as 100% since the service would be free

# Details

This website has the breakdown of which counties the 4 programs serve: https://housing.nv.gov/programs/Weatherization/
From the website:

- [Community Services Agency (CSA)](https://www.csareno.org/programs/?Energy%20Efficiency%20(Weatherization)) service area includes Washoe County - Phone number: (775) 786-6023
- [HELP of Southern Nevada (HELP)](https://www.helpsonv.org/weatherization) service area includes the City of Las Vegas, City of Henderson and Southern Rural Clark County - Phone Number: (702) 795-0575
- [Nevada Rural Housing Authority (NRHA)](https://nvrural.org/weatherization/) service area includes Carson City, Churchill County, Douglas County, Lyon County, City of North Las Vegas, Northern Rural Clark County and Storey County - Phone number: Northern: (775) 887-1795, ext. 112 Southern: (702) 660-5136
- [Rural Nevada Development Corporation (RNDC)](http://www.rndcnv.org/housing-services/weatherization/) service area includes the following counties: Elko, Esmeralda, Eureka, Humboldt, Lander, Lincoln, Mineral, Nye, Pershing, and White Pine Phone Numbers: (775) 289-8519 or toll free (866) 404-5204


data/geo_groups.json
```diff
@@ -35,6 +35,49 @@
       ]
     }
   },
+  "NV": {
```
rpartridge, line 38: none of these overlap except that there are 2 that serve Clark county - HELP serves Southern Rural Clark County and NRHA serves Northern Rural Clark County. I added clark county to both geogroups but that may not be the right approach. I suspect this is an edge case because the majority of people who live in clark county are las vegas, north las vegas, or henderson, which are listed specifically under cities. it might be better in that case to remove clark county from the list to avoid showing incorrect information -- what do people think?

oyamauchi: The semantics of having multiple authority types in a group is "AND" -- i.e. "you must be in one of these cities AND in one of these counties". So `help-of-southern-nevada` is overly restrictive: it will only match if you're within Vegas or Henderson, and the county part is redundant.

`nevada-rural-housing` will only match if you're in North Las Vegas (and thus automatically in Henderson County). Carson City is a county-equivalent, so it will never match because there isn't an object under `counties` with `"county": "Carson City"`. (The `county` key in the objects under `cities` is meaningless; it's never read. We could stand to have a tighter schema on authorities.json, it looks like.)

I'm pretty sure the current semantics of geo groups don't support these cases fully. Do we need the ability to do negation here? Like "Henderson County AND NOT [these cities]"? (I think there's a case in Colorado that would benefit from this too, fwiw.)

src/data/low_income_thresholds.ts
```diff
@@ -31,6 +31,13 @@ export enum ILIncomeAuthority {
   STATE = 'il-state-of-illinois',
 }
 
+export enum NVLowIncomeAuthority {
+  CSA_REN0 = 'nv-csa-reno',
```
oyamauchi, line 35: Why is this a zero?

oyamauchi (changes requested):
I think geo groups need to get more sophisticated to support this properly.

data/geo_groups.json
oyamauchi: BTW, you may be wondering: why is it AND semantics, that seems nonsensical where cities and counties are concerned?

I was mostly thinking about combining either a cities or counties restriction with a utilities restriction, like "customers of XYZ Utility who live in City of ABC" or whatever.  I also realized you could simulate OR semantics by having two different geo groups (one for each side of the OR) and copying the incentive, with each copy having one of the two geo groups.

I didn't account for NOT semantics in any way, which I think is called for here.

coopr:
> I think geo groups need to get more sophisticated to support this properly.

Just wanted to hop in and say how much I feel for ya'll working through these details. 

Here at https://www.QuitCarbon.com we've been working through many of the same details - and it ain't easy! 

We are likely ending up at some similar places - great minds think alike? 

Keep up the great work!

src/data/low_income_thresholds.ts
rpartridge: oops, typo

data/geo_groups.json
rpartridge: ah, good to know that this is AND -- I think there's an okay solution that we can get to here for `help-of-southern-nevada` and `nevada-rural-housing-north`. we could:
1.  update `help-of-southern-nevada` to only include the cities of LV and Henderson and remove clark county
2.  split `nevada-rural-housing-north` into 2 groups/incentives to create an OR that applies to city of north las vegas OR any of the counties

this should cover the whole state except for rural clark county, because clark county won't be included in any of these, otherwise we'd need to support the NOT functionality. I actually think that's okay because northern clark county is served by a different agency than southern clark county (but the website doesn't specify what north and south refer to specifically). since the main population centers in clark county are las vegas, north las vegas, and henderson, that should cover most people, and we would just have to exclude rural clark county for now (which seems acceptable?). alternatively we could just combine all of these into 1 state-wide program and leave it up to users to figure out which one to apply to, but the problem with that is that each of the 4 programs has different low income thresholds. what do you think?

something like:
```
"nv-group-help-of-southern-nevada": {
  "cities": [
    "nv-city-of-las-vegas",
    "nv-city-of-henderson"
  ]
},
"nv-group-nevada-rural-housing-north-lv": {
  "cities": [
    "nv-city-of-north-las-vegas"
  ]
},
"nv-group-nevada-rural-housing": {
  "counties": [
    "nv-churchill-county",
    "nv-douglas-county",
    "nv-lyon-county",
    "nv-storey-county",
    "nv-carson-city"
  ]
},
```

rpartridge: updated in the[ most recent commit](https://github.com/rewiringamerica/api.rewiringamerica.org/pull/345/commits/af002e3cd0f164c1229651c9f06ad65df669a484) and tested with zip codes in las vegas, north las vegas, henderson, and carson city!

oyamauchi: OK, this seems fine -- I also can't find anywhere that defines what "northern rural Clark County" is, so I'm OK with punting on it.

Also, why would they include North Las Vegas in "Nevada Rural Housing"? It's very much not rural!

data/NV/incentives.json
```diff
@@ -1558,6 +1558,98 @@
     },
     "bonus_available": true
   },
+  {
+    "id": "NV-74",
+    "authority_type": "other",
+    "authority": "nv-csa-reno",
+    "eligible_geo_group": "nv-group-csa-reno",
+    "payment_methods": [
+      "pos_rebate"
```
oyamauchi, line 1567: We categorize stuff like this as `assistance_program`, since it's more of a "you get free stuff" situation rather than a "you get all the money back" situation.

data/low_income_thresholds.json
```diff
@@ -286,6 +286,71 @@
...
+      "source_url": "https://static1.squarespace.com/static/5e91c64d83735c5db90418a6/t/65dd07f61363a972f9bc5e65/1708984310854/WEBSITES+INCOME+SHEET+-+2024+WX.pdf",
+      "thresholds": {
+        "1": 30120,
+        "2": 40880.04,
+        "3": 51639.96,
+        "4": 62400,
+        "5": 73160.04,
+        "6": 83919.96,
+        "7": 94680,
+        "8": 105440
```
oyamauchi, line 352: I'm pretty sure this is just rounding error -- e.g. the source PDF lists the monthly limit for 3 people as 4303.33. That probably comes from dividing 51640 / 12, which is 4303.33.... ([repeating of course](https://youtu.be/mLyOj_QD4a4?t=68)). So I think we should just round these to the nearest whole dollar. I've never seen fractional income limits anywhere else.

oyamauchi (approved):

datu925 (approved):

PR #350, opened 2024-03-12
Switch to configuring low-income explicitly by record.

datu925:
This isn't fully complete, but I wanted to get initial feedback before writing tests.

After doing some investigation, I didn't want to go the geo_groups route for low_income configuration for a couple reasons:
1. low-income thresholds are of interest to external stakeholders, and I think it will complicate things for all parties if at some point in the spreadsheet-to-json process, we're replacing data that is standalone and human-readable with an enum key that only makes sense if referencing a separate dataset (this reasoning applies to geo groups too but is maybe less important there)
2. in the spreadsheets, data collectors tend to put notes in the income column that aren't shared across all records that have the same low-income thresholds. Some of these refer to the non-low-income counterpart incentive, if one exists (see CT and RI for examples). There may be other use cases too. I don't want to collapse these notes into the same JSON object like we did with geo_groups.

What I favor instead is keeping low-income in spreadsheet form as a free-text field, and then using config to tie the thresholds to records. I chose records instead of authorities or programs because I don't think it's that much extra config and it's more precise.

While it's redundant to express the records in the thresholds and also have low_income as an incentive field, I think it makes sense since creation of the incentives JSON is automated. In that sense we still have a single representation of low-income information and then the incentives JSON is downstream of that.

If this is okay, I'll add static tests to verify that the info in the thresholds and JSON reflect the same reality.

I'm also getting rid of the default requirement since we won't need that in the future.

### DIFF
--- a/test/data/low_income_thresholds.test.ts
+++ b/test/data/low_income_thresholds.test.ts
@@ -1,0 +1,47 @@
+import { test } from 'tap';
+import { LOW_INCOME_THRESHOLDS_BY_AUTHORITY } from '../../src/data/low_income_thresholds';
+import {
+  STATE_INCENTIVES_BY_STATE,
+  StateIncentive,
+} from '../../src/data/state_incentives';
+
+function computeIdToIncentiveMap(incentives: StateIncentive[]) {
+  const output: { [index: string]: StateIncentive } = {};
+  for (const incentive of incentives) {
+    output[incentive.id] = incentive;
+  }
+  return output;
+}
+
+test('low-income thresholds are equivalent in JSON config and incentives', async t => {
+  // All incentives have the threshold indicated in their config.
+  const map = computeIdToIncentiveMap(
+    Object.values(STATE_INCENTIVES_BY_STATE).flat(),
+  );
+  Object.entries(LOW_INCOME_THRESHOLDS_BY_AUTHORITY).forEach(
+    ([, stateThresholds]) => {
+      for (const [identifier, thresholds] of Object.entries(stateThresholds)) {
+        for (const incentiveId of thresholds.incentives) {
+          const incentive = map[incentiveId];
+          t.equal(incentive.low_income, identifier);
+        }
+      }
+    },
+  );
+  // Converse: all threshold configs accurately reflect the incentive IDs.
+  for (const [stateId, stateIncentives] of Object.entries(
+    STATE_INCENTIVES_BY_STATE,
+  )) {
+    for (const incentive of stateIncentives) {
+      if (incentive.low_income) {
+        const stateThresholds = LOW_INCOME_THRESHOLDS_BY_AUTHORITY[stateId];
+        t.hasProp(stateThresholds, incentive.low_income);
+        t.ok(
+          stateThresholds[incentive.low_income].incentives.includes(
+            incentive.id,
+          ),
+        );
+      }
+    }
+  }
+});
--- a/data/low_income_thresholds.json
+++ b/data/low_income_thresholds.json
@@ -1,7 +1,7 @@
 {
   "AZ": {
     "default": {
-      "records": [
+      "incentives": [
         "AZ-18",
         "AZ-21"
       ],
@@ -20,7 +20,7 @@
   },
   "CO": {
     "co-black-hills-energy": {
-      "records": [
+      "incentives": [
         "CO-24",
         "CO-25",
         "CO-26"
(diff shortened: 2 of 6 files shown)


oyamauchi (commented):
I think I get what's happening here; some notes below.

- Is the idea that `incentives-spreadsheet-to-json`, when writing out each incentive, would look for its ID in low income thresholds to populate the `low_income` field? (And ignore the income eligibility info in the spreadsheet)

- I'm on the fence re: whether the `low_income` field in incentives is necessary. As you note, it's redundant. It wouldn't be hard to switch to using the config from `low_income_thresholds` during calculation. But then if you were looking at just an incentive object in JSON, you wouldn't know whether it has an income restriction; there'd be "action at a distance" happening. I could really go either way, unless there's some pro or con I'm not thinking of.

- Do you think we should do this for geo groups too? I think I'm OK with it, and now would be the time to do it (since there are no geographic-restriction columns in the sheets yet). The upside is a consistent pattern/workflow between income and geo restrictions, which I think will become more important as we get more people working in this data. (I want to avoid having to say things to new people like "well _this_ part is done differently because of weird historical reasons", you know?)

- I'm in agreement with having the association be at the incentive level, rather than program or authority; otherwise we start overloading the latter concepts again.

- [bikeshed] I would call the field `incentives` or `incentive_ids`, not `records`; "record" is a pretty generic term and we're not using it anywhere in the API.

datu925:
> I think I get what's happening here; some notes below.
> 
> * Is the idea that `incentives-spreadsheet-to-json`, when writing out each incentive, would look for its ID in low income thresholds to populate the `low_income` field? (And ignore the income eligibility info in the spreadsheet)

That's right – the association is done using whatever is in the low income thresholds file rather than parsing the column like we do for geo groups. The script can still raise "possible dumb error" warnings if an incentive has the low_income free-text field populated in the spreadsheet but nothing configured in the thresholds file. 

> * I'm on the fence re: whether the `low_income` field in incentives is necessary. As you note, it's redundant. It wouldn't be hard to switch to using the config from `low_income_thresholds` during calculation. But then if you were looking at just an incentive object in JSON, you wouldn't know whether it has an income restriction; there'd be "action at a distance" happening. I could really go either way, unless there's some pro or con I'm not thinking of.

I'm also torn. We have a precedent for "action at a distance" with the incentive relationships, but I don't actually love that aspect of it. It's nice if the incentive record has all the context.

I'm inclined to keep it on the record since it should be automated anyway, but if it becomes a pain point in the future, I wouldn't hesitate to take it off.

> * Do you think we should do this for geo groups too? I think I'm OK with it, and now would be the time to do it (since there are no geographic-restriction columns in the sheets yet). The upside is a consistent pattern/workflow between income and geo restrictions, which I think will become more important as we get more people working in this data. (I want to avoid having to say things to new people like "well _this_ part is done differently because of weird historical reasons", you know?)

Agree, best to have a unified approach if possible. I personally favor making the switch – I like this approach a bit better for the reasons I mentioned in this PR's intro comment.

Happy to do as a follow-up, but let me know if you want to do it.

> * I'm in agreement with having the association be at the incentive level, rather than program or authority; otherwise we start overloading the latter concepts again.

Ack, great.

> * [bikeshed] I would call the field `incentives` or `incentive_ids`, not `records`; "record" is a pretty generic term and we're not using it anywhere in the API.

Good call, will fix.

oyamauchi:
OK, then I'm fine with this approach and keeping the `low_income` key in incentive records.

I think you may as well do the switching of geo groups too as a followup, since the structure of it will be basically the same as this.

datu925:
Changes:
- rename from `records` to `incentives`
- static tests to verify equivalence of config and incentives
- spreadsheet-to-json script support

test/data/low_income_thresholds.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  // All incentives have the threshold indicated in their config.
+  const map = computeIdToIncentiveMap(
+    Object.values(STATE_INCENTIVES_BY_STATE).flat(),
+  );
+  Object.entries(LOW_INCOME_THRESHOLDS_BY_AUTHORITY).forEach(
+    ([, stateThresholds]) => {
+      for (const [identifier, thresholds] of Object.entries(stateThresholds)) {
+        for (const incentiveId of thresholds.incentives) {
+          const incentive = map[incentiveId];
+          t.equal(incentive.low_income, identifier);
```
oyamauchi, line 26: This would catch a case where multiple threshold configs point to the same incentive, right?

oyamauchi (approved):

PR #351, opened 2024-03-12
Adding Spanish descriptions for CO

mariabajzek:
**Test Plan**

I mostly just checked that punctuation looked ok and that the English ones weren't negatively affected in the process of running the script.

### DIFF
--- a/data/CO/incentives.json
+++ b/data/CO/incentives.json
@@ -2210,7 +2210,7 @@
     ],
     "short_description": {
       "en": "Rebate of 25% of the purchase price (up to $250) off a corded or battery powered 2 stage Snow Blower.",
-      "es": "Un reembolso del 25% del precio de compra (hasta $250 dólares) en un soplador de nieve de 2 etapas con cable o batería."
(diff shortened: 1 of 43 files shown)


datu925:
Would you mind doing the following?
1. Removing 'CO' from line 15 of scripts/spreadsheets-health.test.ts
2. Running yarn check-spreadsheet-health and confirming that CO passes

mariabajzek:
Well... it doesn't pass, but it also doesn't pass on main I think because of a missing authority for NV.

mariabajzek:
Ah, it's some missing programs actually:

nv_cSAReno_weatherizationAssistanceProgram
nv_ruralNevadaDevelopmentCorporation_weatherizationAssistanceProgram
nv_nevadaRuralHousing_weatherizationAssistanceProgram
nv_hELPOfSouthernNevada_weatherizationAssistanceProgram

I'll see if I can add them.

mariabajzek:
Hm ok I don't think I have time to fix the Nevada stuff since I don't think just adding the programs would bring it in sync with the spreadsheet. But since the errors are not related to CO I think this is ok for this PR.

datu925 (approved):
I only wanted to make sure CO passed – not concerned about the other states. Thanks.

PR #352, opened 2024-03-12
Turn start/end dates into strings

oyamauchi:
## Description

This is mechanical: just stringifying the `start_date` and `end_date`
of every incentive record, and changing the v1 API to serve them as
strings instead of numbers.

In the process, I made the slightly horrifying discovery that `t.same`
in Tap considers values to be "same" up to type coercion; that is,
`2024` and `"2024"` are considered "same". I changed usages of that
test method to `strictSame`, which doesn't do type coercion.

I'll follow up with a PR to be smarter about parsing the values that
are in spreadsheets -- some state sheets have precise dates.

The new embed frontend has already been updated to handle the dates
being either numbers or strings. (It currently doesn't even read them,
and in fact the server code is hard-overwriting all state incentives'
start/end dates with 2023 and 2024 respectively. I'll be changing the
latter shortly.)

## Test Plan

`yarn lint` and `yarn test`. New test to validate the regex I'm using
to enforce date format in the JSON files.

### DIFF
--- a/data/AZ/incentives.json
+++ b/data/AZ/incentives.json
@@ -109,7 +109,7 @@
       "en": "Rebate of $500 for each minimum usable capacity battery unit of 5 kWh, installed on or after Jan 9, 2022.",
       "es": "Un reembolso de $500 dólares por cada unidad de batería de 5 kWh de capacidad mínima utilizable instalada a partir del 9 de enero del 2022."
     },
-    "start_date": 2022
+    "start_date": "2022"
   },
   {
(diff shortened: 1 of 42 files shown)


datu925 (approved):
Looks good – I think this shouldn't cause any new failures in the spreadsheet-health-tests, but if you could run `yarn check-spreadsheet-health` to make sure there aren't any date-related failures, that would be great (apparently Nevada is failing currently so you can ignore that).

veekas (approved):

PR #355, opened 2024-03-13
Update geo groups to be configured similar to low-income thresholds.

datu925:
The implementation in refineCollectedData is a little repetitive, but I didn't see an obvious way to refactor it.

### DIFF
--- a/scripts/lib/data-refiner.ts
+++ b/scripts/lib/data-refiner.ts
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import { GeoGroupsByState } from '../../src/data/geo_groups';
 import { LowIncomeThresholdsMap } from '../../src/data/low_income_thresholds';
 import {
   CollectedIncentive,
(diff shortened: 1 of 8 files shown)


test/data/schemas.test.ts
```diff
@@ -206,25 +205,6 @@ test('state incentives JSON files match schemas', async tap => {
...
-          incentive.eligible_geo_group!,
-          `nonexistent geo_group (${stateId}, id ${incentive.id})`,
-        );
-      } else {
-        tap.notOk(
-          incentive.eligible_geo_group,
-          `only authority_type 'other' can have a geo group (id ${incentive.id})`,
-        );
-      }
-
```
datu925, line 227: Just flagging that I moved these into the new test file to keep geogroup-related tests together. The middle test (nonexistent group ID) is implicitly covered by the test that verifies equivalence of config and incentives.

scripts/lib/data-refiner.ts
```diff
@@ -135,15 +151,6 @@ function isPlausibleLowIncomeRow(record: CollectedIncentive) {
   return false;
 }
 
-/** An ID-like string that includes "-group-", like "nv-group-some-counties". */
-function isPlausibleGeoGroup(geo_eligibility: string | undefined) {
```
oyamauchi, line 139: Yeah, seeing it now, I like that the `-group-` thing is no longer load-bearing.

oyamauchi (approved):
This will conflict with #345 , so could you wait until that's landed and fix up the NV geo groups file before landing this? (Or coordinate with Rachel to do it in the opposite order)

PR #356, opened 2024-03-13
Try to parse full dates from spreadsheets; make dates optional in v1

oyamauchi:
## Description

We'll want to change the spreadsheets' guidance for how to input
dates, but for now, it's easy enough to parse all the date formats
that people have input.

I chose this datetime library because it's designed to resemble
`java.time`, which is the most sensible datetime library I've used.

Separately, it's now pretty clear to me that it doesn't make sense to
have the date fields be required in the v1 API -- most state
incentives don't have dates defined, and so far we've just been
hardcoding in `start_date: 2023` and `end_date: 2024` even for
incentives that do define their own dates. (Fortunately the frontend
isn't even reading them.) So they're now optional.

## Test Plan

`yarn lint` and `yarn test`.

I ran the spreadsheet-to-JSON script on AZ, CO, and VT, which are the
only ones that have any incentives with dates in them, and made sure
there were no failures due to unparseable dates. AZ and CO succeeded;
changes are reflected here. VT failed for unrelated reasons; will look
into that separately.

Added an unparseable date to an incentive in the CO sheet and made
sure the script output that row as invalid.


scripts/lib/data-refiner.ts
```diff
@@ -144,20 +147,47 @@ function isPlausibleGeoGroup(geo_eligibility: string | undefined) {
...
-  } else if (input.indexOf('-') !== -1) {
-    sep = '-';
-  } else {
+/**
+ * This is for start and end dates of incentives. If the input is already in
+ * the required date format, it's returned as-is.
+ *
+ * Otherwise, we make an effort to parse a date from a few common formats.
+ * Ideally, though, spreadsheets should use our incentive date format directly
+ * (a superset of ISO 8601 dates).
```
datu925, line 156: I think this is a good follow-up PR. If I'm understanding correctly, ajv is capable of validating using that regex pattern. If so, we can get rid of start_date_raw and end_date_raw and just make both date fields pass-through, and it's a CollectedIncentive schema validation error if the spreadsheet has an invalid date format.

datu925 (commented):
Won't this change cause the spreadsheets and JSON to go out of sync for other states as well? For instance, GA has a bunch of 12/31/2025 dates, and if I'm tracking here, those should be appearing as 2025-12-31.

oyamauchi:
Somehow I missed that a bunch of other states had dates populated, even though I wrote a little script to check them... weird.

I think the followup work is (a) update the spreadsheet template to require the pseudo-ISO-8601 date format, (b) change all the dates in existing spreadsheets, which is going to be a pain, and (c) update the script to require the pseudo-ISO-8601 format.

datu925:
> Somehow I missed that a bunch of other states had dates populated, even though I wrote a little script to check them... weird.
> 
> I think the followup work is (a) update the spreadsheet template to require the pseudo-ISO-8601 date format, (b) change all the dates in existing spreadsheets, which is going to be a pain, and (c) update the script to require the pseudo-ISO-8601 format.

It might make sense to do that just to get it over with, but if you punt for a bit, the eventual json-back-to-spreadsheet flow will take care of it.

If so, you could:
1. move the date standardization you've written here from data-refiner.ts into spreadsheet-standardizer.ts. That way the dates are converted on their way into the CollectedIncentive schema, and it's an ajv schema error if they're in a format that is unconvertable. If they're convertable, they don't need to be changed.
2. once we have a working json-back-to-spreadsheet flow, then running spreadsheet-to-json will make the date canonical and then json-to-spreadsheet will write the canonical date back to spreadsheet, letting us replace the old sheets.

datu925 (commented):
What are the scripts/look.ts and scripts/change.ts files? Are those supposed to be here?

datu925 (approved):

PR #359, opened 2024-03-14
Add exclusion relationships to WI incentives.

hblumberg:
This PR handles the three incentives that have non-restricted and income-qualified versions.

## Links
- _Asana_ https://app.asana.com/0/1206629812394927/1206842549215196/f

## Checklist

## Test Plan
* `yarn test`


scripts/README.md
```diff
@@ -60,6 +60,20 @@ All valid records according to our schema will be written to the location you sp
 Note that even after running these, there are still files you must edit manually to get the JSON in. Follow a recent CL example.
 Eg: https://github.com/rewiringamerica/api.rewiringamerica.org/pull/209/files
 
+### Incentive Relationships
```
oyamauchi, line 63: This stuff is also documented in `docs/relationships-README.md`, which I get is not super discoverable. It's also a bit more abstract and not focused on the sheet-to-JSON process.

Could you either fold this info into that doc, or the other way around, or whatever you think is best?

oyamauchi (approved):
The actual change looks fine but we need to sort out the docs thing.

scripts/README.md
hblumberg: Oops, I totally missed that -- thank you! What do you think of linking to that README here?

oyamauchi: I think that's fine -- `docs` can be for more in-depth stuff, and this file for practical sheet-to-JSON info.

PR #360, opened 2024-03-14
Add OR incentives

hblumberg:
Note that this work did NOT include a data refresh.

## Links
- _Asana_ https://app.asana.com/0/1206629812394927/1206778652392985/f

## Checklist

## Test Plan
* Ensure `node build/scripts/incentive-spreadsheet-to-json.js --strict OR` runs without error
* Ensure `yarn test` runs without error


data/OR/incentives.json
```diff
@@ -0,0 +1,962 @@
...
+    "authority_type": "utility",
+    "authority": "or-portland-general-electric",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "rooftop_solar_installation",
+    "program": "or_portlandGeneralElectric_solarWithinReach",
+    "amount": {
+      "type": "dollars_per_unit",
+      "number": 1,
```
datu925, line 273: This should be 0.9. I think it's a formatting issue with the spreadsheet not having sufficient decimal points. Trying fix that so it shows up as 0.9 in the spreadsheet and then rerunning (you might have to wait a few minutes for it to update).

data/OR/incentives.json
```diff
@@ -0,0 +1,962 @@
...
+    "authority_type": "utility",
+    "authority": "or-pacific-power",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "rooftop_solar_installation",
+    "program": "or_pacificPower_solarWithinReach",
+    "amount": {
+      "type": "dollars_per_unit",
+      "number": 1,
```
datu925, line 296: Same issue.

data/OR/incentives.json
```diff
@@ -0,0 +1,962 @@
...
+      "unit": "square_foot"
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$1 per sqft for high efficiency windows. Windows must replace existing windows, glass doors or skylights. U-value must be between 0.27 – 0.23."
```
datu925, line 393: optional: I don't think I've seen `sqft` before – take a look through our other incentives to see if we have a different way of representing this, e.g. `sq. ft`.

If so, change the others like it.

data/OR/incentives.json
```diff
@@ -0,0 +1,962 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 250
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$250 discount on qualifying central air conditioners. Outdoor unit is required to achieve a minimum nominal rating of SEER2 15.20 or greater."
```
datu925, line 659: We don't generally do air conditioners, since they're already electric and we want people to get heat pumps instead.

Add a column called `Omit from API?` to the spreadsheet and put a `TRUE` value there for this row. When you rerun the script, it should not have this incentive anymore.

Alternatively, you could delete it entirely, though I suppose it might be nice to keep around in case we change our minds.

data/OR/incentives.json
```diff
@@ -0,0 +1,962 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 100
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$100 discount on qualifying central air conditioners. Outdoor unit is required to achieve a minimum nominal SEER2 rating between 14.25 – 15.19."
```
datu925, line 680: Same as above.

data/OR/incentives.json
```diff
@@ -0,0 +1,962 @@
...
+    ],
+    "item": "heat_pump_air_conditioner_heater",
+    "program": "or_energyTrustOfOregon_energyTrustOfOregon",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 2000
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
```
datu925, line 38: I think the `owner_status` for OR is frequently wonky. A portable heat pump maybe, but ductless *replacing electric resistance* is not something a renter can do.

It's a bit of a pain, but I think you can quickly redo this column in the spreadsheet by looking at the list in https://github.com/rewiringamerica/incentive-data-llm/blob/d56ac40232df4fb5f6f529584f82988e07383fef/src/constants.ts#L16. That has a fairly reliable mapping between the Technology type and owner_status (though it's a bit old so some Technologies may be missing).

I think it'll take ~15 minutes to do by hand, sorry.

datu925 (changes requested):
Some spreadsheet issues, but not too many this time. Otherwise looks good.

data/OR/incentives.json
hblumberg: Ah okay, will do! I think they were coded this way because EnergyTrust specifies the eligible residence types (Owner Occupied vs. Rental Property). But even if a rental property is eligible, I agree that the renter won't be the decision maker.

If the technologies map to the owner status, could we just remove the owner status from the spreadsheets? (Out of scope for this PR of course, just wondering.)

datu925: It's a good question. I feel like this came up once before but I can't remember when/in what forum.

I think it's pretty reasonable to make these properties of the underlying Item/Technology and take them off the individual incentives, but I'll defer to @oyamauchi and @RandomEtc.

data/OR/incentives.json
```diff
@@ -0,0 +1,892 @@
...
+    "program": "or_pacificPower_pacificPowerRebates",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 500
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Up to $500 rebate for qualifying hardwired level 2 chargers, capped at 75% of total costs. Must be installed by a licensed electrician."
```
datu925, line 683: For OR-34 through OR-37, these are better represented as amount type percent (with amount numbers as 0.75), and the current dollar values should become amount maximums. The descriptions don't need to change.

datu925 (changes requested):
Last critique!

PR #361, opened 2024-03-14
v0: add new more_info_url_internal field to expose new federal IRA incentive pages

rpartridge:
Asana ticket: https://app.asana.com/0/1206540452900052/1206673169101907/f

## What

The consumer app team is releasing a bunch of new pages with information on the IRA incentives. We want to update the federal IRA incentive "learn more" links to point to these new pages.

Note that none of these pages are live yet and won't be until Monday.

## Releasing

This PR is safe to be released today because it's only modifying the data in `ira_incentives.json` and adding a new field to v0 to expose these links. It shouldn't make any changes to v1 or to the existing v0 clients.

### DIFF
--- a/src/data/ira_incentives.ts
+++ b/src/data/ira_incentives.ts
@@ -34,6 +34,7 @@ export interface IRAIncentive {
     | PaymentMethod.PerformanceRebate;
   payment_methods: PaymentMethod[];
   short_description: LocalizableString;
+  url?: LocalizableString;
 }
 
 // Work around https://github.com/ajv-validator/ajv/issues/1664
@@ -91,6 +92,9 @@ export const SCHEMA: JSONSchemaType<IRAIncentive[]> = {
       },
       short_description: {
         $ref: 'LocalizableString',
+      },
+      url: {
+        $ref: 'LocalizableString'
       },
     },
     required: [
--- a/src/data/ira_incentives.ts
+++ b/src/data/ira_incentives.ts
@@ -34,7 +34,7 @@ export interface IRAIncentive {
     | PaymentMethod.PerformanceRebate;
   payment_methods: PaymentMethod[];
   short_description: LocalizableString;
-  url?: LocalizableString;
+  more_info_url?: LocalizableString;
 }
 
 // Work around https://github.com/ajv-validator/ajv/issues/1664
@@ -93,7 +93,7 @@ export const SCHEMA: JSONSchemaType<IRAIncentive[]> = {
       short_description: {
         $ref: 'LocalizableString',
       },
-      url: {
+      more_info_url: {
         $ref: 'LocalizableString',
       },
     },
--- a/src/data/ira_incentives.ts
+++ b/src/data/ira_incentives.ts
@@ -94,7 +94,7 @@ export const SCHEMA: JSONSchemaType<IRAIncentive[]> = {
         $ref: 'LocalizableString',
       },
       url: {
-        $ref: 'LocalizableString'
+        $ref: 'LocalizableString',
       },
     },
     required: [
--- a/src/data/ira_incentives.ts
+++ b/src/data/ira_incentives.ts
@@ -34,7 +34,7 @@ export interface IRAIncentive {
     | PaymentMethod.PerformanceRebate;
   payment_methods: PaymentMethod[];
   short_description: LocalizableString;
-  more_info_url?: LocalizableString;
+  more_info_url: LocalizableString;
 }
 
 // Work around https://github.com/ajv-validator/ajv/issues/1664
--- a/src/routes/v0.ts
+++ b/src/routes/v0.ts
@@ -42,6 +42,10 @@ function translateIncentives(incentives: IRAIncentive[]): WebsiteIncentive[] {
       'https://www.rewiringamerica.org',
       '',
     ),
+    // TODO: remove when PEP is migrated to v1 API
+    more_info_url_internal:
+      incentive.more_info_url?.en ||
+      tr(PROGRAMS[incentive.program as keyof Programs].name, 'en'),
     short_description: tr(incentive.short_description, 'en'),
 
     type:
--- a/src/routes/v0.ts
+++ b/src/routes/v0.ts
@@ -43,9 +43,7 @@ function translateIncentives(incentives: IRAIncentive[]): WebsiteIncentive[] {
       '',
     ),
     // TODO: remove when PEP is migrated to v1 API
-    more_info_url_internal:
-      incentive.more_info_url?.en ||
-      tr(PROGRAMS[incentive.program as keyof Programs].name, 'en'),
+    more_info_url_internal: tr(incentive.more_info_url, 'en'),
     short_description: tr(incentive.short_description, 'en'),
 
     type:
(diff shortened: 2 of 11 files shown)


test/fixtures/v0-80212-homeowner-80000-joint-4.json
```diff
@@ -13,7 +13,7 @@
       "program_es": "Reembolsos de Electrificación y Electrodomésticos (HEAR)",
       "item": "Heat Pump Air Conditioner/Heater",
       "item_es": "Aire acondicionado/calentador con bomba de calor",
-      "more_info_url": "/app/ira-calculator/information/heat-pump-air-conditioner-heater",
+      "more_info_url": "https://homes.rewiringamerica.org/federal-incentives/25C-heat-pump-tax-credits",
```
oyamauchi, line 16: If I'm reading the www code correctly, this will still work there, but I can imagine an API v0 client assuming that these URLs will always be relative and always prepending `https://www.rewiringamerica.org` to absolutize it.

As far as I know, we haven't formally documented anywhere that these URLs are relative, but I think this is a breaking change in spirit if not in letter. @RandomEtc what do you think? We should probably send an email to API clients either way, but should we block on this?

locales/en.json
```diff
@@ -1,19 +1,19 @@
...
-    "geothermal_heating_installation": "https://www.rewiringamerica.org/app/ira-calculator/information/geothermal-heating-installation",
-    "heat_pump_air_conditioner_heater": "https://www.rewiringamerica.org/app/ira-calculator/information/heat-pump-air-conditioner-heater",
-    "heat_pump_clothes_dryer": "https://www.rewiringamerica.org/app/ira-calculator/information/heat-pump-clothes-dryer",
-    "heat_pump_water_heater": "https://www.rewiringamerica.org/app/ira-calculator/information/heat-pump-water-heater",
-    "new_electric_vehicle": "https://www.rewiringamerica.org/app/ira-calculator/information/electric-vehicles",
-    "rooftop_solar_installation": "https://www.rewiringamerica.org/app/ira-calculator/information/rooftop-solar-installation",
-    "used_electric_vehicle": "https://www.rewiringamerica.org/app/ira-calculator/information/electric-vehicles",
-    "weatherization": "https://www.rewiringamerica.org/app/ira-calculator/information/weatherization"
+    "battery_storage_installation": "https://homes.rewiringamerica.org/federal-incentives/25d-battery-storage-tax-credit",
```
oyamauchi, line 3: I think these make more sense as URLs on the IRA programs, not on items. You'd add them in `src/data/programs.ts`. Then you'll also have to edit `src/routes/v0.ts` to populate `more_info_url` with the program URL, rather than the item URL.

Then these item URLs would become sort of useless. They wouldn't appear at all in v0. They'd still be there in v1, but the frontend doesn't use them except as a fallback if there's no `program_url` (and I think all non-IRA programs have URLs now). As a followup, we should make `url` required on Program objects, make `program_url` required in v1, and reflect that in the frontend code.  And possibly get rid of these item URLs altogether?

oyamauchi (changes requested):

test/fixtures/v0-80212-homeowner-80000-joint-4.json
rpartridge: I'd propose that we just leave the v0 urls the same then and only update these for v1. it sounds like we can do that with your suggestion below of making these program URLs, and then returning the program URLs for v1, but continue using item urls for v0. I'm going to start working on that but let me know if that doesn't make sense

oyamauchi: Yeah, that seems fine. It hadn't quite registered for me that this is all happening in conjunction with shutting down the old calculator anyway, so we don't need to have the new URLs in v0.

rpartridge: hmm, this approach isn't right either -- the new pages are really per incentive, not per program. a single IRA program could have many links associated with different incentives. incentive_item actually also isn't the right level of granularity, because one incentive item could have 2 links (one for a rebate, one for a tax credit). I think these will need to be per incentive

rpartridge:
@oyamauchi I ended up going with a different approach because the new links that need to be added are too granular to be either program urls or incentive item URLs. for example, there are different links for heat pump rebates vs heat pump tax credits, so we can't have a single link for the program or heat pump incentive item.

to deal with this, I added some urls to programs, for the 4 programs that only have 1 url (EV charger, new EV, used EV, efficiency rebates). for the remaining incentives, I added a new field `url` to the IRA incentives that links to that specific incentive's page. the Spanish links are still the legacy rewiringamerica.org ones, but the Spanish translations of the new pages will be done in ~2 weeks and can be updated then

test/fixtures/v0-80212-homeowner-80000-joint-4.json
RandomEtc: I’m not totally following the mapping from old to new URLs but yes I would consider moving to an absolute URL a breaking change here. There definitely are API v0 customers who display those links (though admittedly not many, so yes it might be worth contacting them). 

I like the idea of adding a new URL field for v0 to avoid it being a breaking change, and then making sure this all makes sense in v1 before we release that publicly.

oyamauchi (approved):
The core of the problem is that there's no other incentives that are part of multi-item programs where we want separate URLs for each item. This is a new concept in the API and data.

One possible alternative to what you have here is to split the programs in `programs.ts` -- for example, make one copy of `energyEfficientHomeImprovementCredit` (i.e. 25C) per item that it covers, let them all have the same user-visible name, and different URLs. That's not great because it slightly abuses the concept of a "program" -- there really is only one 25C program in reality.

All in all I think I prefer the approach you have here, but I'd like to **change the name of the field in the data to `more_info_url`**, just to make it a bit clearer what it is. (The fact that it's being served as item.url in v1 is not ideal, but I think not worth fixing at the moment -- I'd rather fix that as part of getting rid of the concept of item URLs entirely.)

src/data/ira_incentives.ts
```diff
@@ -34,6 +34,7 @@ export interface IRAIncentive {
     | PaymentMethod.PerformanceRebate;
   payment_methods: PaymentMethod[];
   short_description: LocalizableString;
+  more_info_url?: LocalizableString;
```
oyamauchi, line 37: I think you can make this required, right?

src/routes/v0.ts
```diff
@@ -42,6 +42,10 @@ function translateIncentives(incentives: IRAIncentive[]): WebsiteIncentive[] {
       'https://www.rewiringamerica.org',
       '',
     ),
+    // TODO: remove when PEP is migrated to v1 API
+    more_info_url_internal:
+      incentive.more_info_url?.en ||
+      tr(PROGRAMS[incentive.program as keyof Programs].name, 'en'),
```
oyamauchi, line 48: `incentive.more_info_url` should also be passed to `tr`, right? (I know this is supposed to be temporary but I could imagine the new pages being translated into Spanish before we're ready to move PEP off of v0)

And if you make more_info_url required, you can remove the fallback. (Also the fallback isn't a URL?)

oyamauchi (commented):
Leaving my previous approval up in case I'm misunderstanding something, but the v0 part looks a little weird to me.

src/routes/v0.ts
rpartridge: PEP doesn't support translations so I didn't think it would matter since we would never send with `language: es` param - but will update just in case

PR #362, opened 2024-03-14
Add basic website scraping, pdf checking, and hashing functions.

mmezher:
Add functions and tests for each function.


test/scripts/website-comparisons.test.ts
```diff
@@ -0,0 +1,71 @@
...
+test('Only HTML body and readable content is returned', tap => {
+  const unclean_html: string =
+    "<html> <head> <meta content='Incorrect to be shown'> <link as='style' href='/doc/assets/stylesheets/fonts-b49e8aae.css' rel='stylesheet'> <meta content='ie=edge' http-equiv='X-UA-Compatible'> </head> <body>Correct content<div> and div contents</div> </body> </html>";
+  tap.equal(cleanWebsiteData(unclean_html), 'Correct content and div contents');
+  tap.end();
+});
+
+test('If no body is present, and no readable html, an empty string is returned', tap => {
+  const unclean_html: string =
+    "<html> <head> <meta content='Incorrect to be shown'> <link as='style' href='/doc/assets/stylesheets/fonts-b49e8aae.css' rel='stylesheet'> <meta content='ie=edge' http-equiv='X-UA-Compatible'> </head></html>";
```
datu925, line 17: Can you add a comment indicating why this is unreadable? Looks fine to my human eyes so it's a big confusing.

test/scripts/website-comparisons.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  tap.equal(cleanWebsiteData(unclean_html), '');
+  tap.end();
+});
+
```
datu925, line 21: Add a test for the no-body-but-readable-html case?

scripts/lib/website-comparisons.ts
```diff
@@ -0,0 +1,46 @@
...
+  return false;
+}
+
+// Given a website link and its html, return a hash of the cleaned website. Do not return a hash if the site is for a PDF.
+export function checkWebsiteAndReturnHash(
+  link: string,
+  response: string,
+): string | null {
+  if (isPdf(link, response)) {
+    return null;
```
datu925, line 37: Out of curiosity, why not just return a hash of the entire PDF content?

This could be naive, but I would guess PDFs change rarely. Doesn't have to happen in this PR either way.

datu925 (approved):
See comments but LGTM.

test/scripts/website-comparisons.test.ts
mmezher: Absolutely, added here.

test/scripts/website-comparisons.test.ts
mmezher: Done

scripts/lib/website-comparisons.ts
mmezher: This is what I'll be aiming to do in a further CL; the issue is that I'm unable to get the PDF content via Cheerio's API - I need to use a PDF parsing library, which has some complexities I'd rather deal with separately.

PR #368, opened 2024-03-15
Make the Energy Star utility data authoritative

oyamauchi:
## Description

Use the Energy Star dataset to generate authorities.json, rather than
having authorities.json hold a hand-curated set of utilities. We can
still manually add logos to utilities in authorities.json, but their
names will come from Energy Star.

The zip-to-utilities table in SQLite is unchanged, but the `utilities`
table is now gone. This changes the initial curation workflow a little
bit -- slightly for the worse in my opinion -- but overall I think
it's still fine. We'll only have to do it once per state.

This will result in us showing a bunch of utilities for which we
haven't done data collection at all, or which we've confirmed do not
offer any incentives. The justification for doing this is that we'll
be much more likely to let users choose their actual utility, rather
than having to pick the wrong answer, or an uninformative "Other"
option. If we decide that we want to be able to distinguish between
these kinds of utilities (still an open product question), we can
easily add a field to the struct in authorities.json to capture those
distinctions.

## Test Plan

`yarn lint` and `yarn test`.

Look at the diff to authorities.json. (There's some churn because I'm
sorting the utilities by ID now, but there shouldn't be such churn in
future.)

Query `/api/v1/utilities` for a sample zip code that was previously
not returning any utility results (89040) and make sure there's a
result now, and that querying `/calculator` with that utility ID
doesn't error. (There won't be any incentives for that utility, but
the point is the API accepts it as a valid utility ID.)


scripts/generate-utility-data.ts
```diff
@@ -256,21 +262,45 @@ enum Col {
...
-  const utilitiesOut = stringify({
-    header: true,
-    columns: ['utility_id', 'name'],
-  });
-  utilitiesOut.pipe(
-    fs.createWriteStream(path.join(__dirname, 'data/utilities.csv')),
+  // Update authorities.json with the utilities from the dataset.
+  const authoritiesJsonPath = path.join(__dirname, '../data/authorities.json');
+  const authoritiesJson: AuthoritiesByState = JSON.parse(
+    await promisify(fs.readFile)(authoritiesJsonPath, 'utf-8'),
```
datu925, line 280: optional – use the promises version of fs instead of promisify. Same with writeFile.

scripts/generate-utility-data.ts
```diff
@@ -135,6 +139,8 @@ const OVERRIDES = new Map<string | number, string>([
   [27316, 'Stowe Electric Department'],
 ]);
 
+type UtilityMap = Map<string, { name: string }>;
+
```
datu925, line 143: I'm not sure if there are style guide considerations advising the contrary, but to me if you're only using this in one method, I'd introduce it there instead.

datu925 (changes requested):
There are some spreadsheet-to-json flows that have to change as a result of this as well. I think they're small enough where you can do them in this PR.

1. generate-misc-state-data.ts should no longer update the authorities file
2. that means the authority-and-program-updater.ts no longer needs the updateAuthoritiesJson() method
3. I do think that generate-misc-state-data.ts should error out if it finds utilities that aren't in the database. Maybe this piece is separate enough to be a follow-up.

oyamauchi:
> generate-misc-state-data.ts should no longer update the authorities file

It still needs to write cities and counties, so I think it should just filter out utilities and otherwise stay the same, right? I'll update with that.

> I do think that generate-misc-state-data.ts should error out if it finds utilities that aren't in the database. Maybe this piece is separate enough to be a follow-up.

There'd be a bit of a circular dependency problem with this. generate-misc-state-data expects authorities.json to have nothing for the state you're updating, so it should be run before generate-utility-data. However, generate-utility-data is the thing that generates the database. There is a unit test (in this PR) that makes sure the set of utilities in authorities.json is the same as the set in the DB, which I think is good enough? I know the unit test is a little downstream of where the problem is, but fixing that would take more work (making generate-misc-state-data non-destructive, basically).

datu925:
> It still needs to write cities and counties, so I think it should just filter out utilities and otherwise stay the same, right? I'll update with that, but let me know if I misunderstood.
> 

You're right, I was only thinking about utility authorities.

> > I do think that generate-misc-state-data.ts should error out if it finds utilities that aren't in the database. Maybe this piece is separate enough to be a follow-up.
> 
> There'd be a bit of a circular dependency problem with this. generate-misc-state-data expects authorities.json to have nothing for the state you're updating, so it should be run before generate-utility-data. However, generate-utility-data is the thing that generates the database. There is a unit test (in this PR) that makes sure the set of utilities in authorities.json is the same as the set in the DB, which I think is good enough? I know the unit test is a little downstream of where the problem is, but fixing that would take more work (making generate-misc-state-data non-destructive, basically).

generate-misc-state-data overwrites the target state's `authorities.json` entry because that was fine historically, but I think with this change, that's no longer appropriate. But making it non-destructive doesn't seem too bad – could we take the existing value of the `utility` key for the state being edited and copy it into the new object before processing the rest of the spreadsheet rows?

Then I think the process would be:
1. generate the new state's authorities with generate-utility-data
2. run generate-misc-state-data. I do think it would be nice to have a check here that an authority in the spreadsheet is found in authorities.json (which implicitly means it's in the csv), but that could be a follow up.

What do you think?

oyamauchi:
OK, I think this reflects that workflow. It goes:

1. Add state key to authorities.json manually (because step 2 only populates states that are already there, to let us choose when to populate utilities)
2. Run `generate-utility-data.ts` to populate utilities for the state
3. Run `generate-misc-state-data.ts` to populate the other auth types, and check that all the utilities in the spreadsheet are in authorities.json already

datu925:
> OK, I think this reflects that workflow. It goes:
> 
> 1. Add state key to authorities.json manually (because step 2 only populates states that are already there, to let us choose when to populate utilities)
> 2. Run `generate-utility-data.ts` to populate utilities for the state
> 3. Run `generate-misc-state-data.ts` to populate the other auth types, and check that all the utilities in the spreadsheet are in authorities.json already

Looks great, thanks. As a follow-up, you could add a command-line arg to `generate-utility-data.ts` that specifies the state you want to add, but it's not like step #1 is particularly onerous as is.

datu925 (approved):

PR #369, opened 2024-03-18
Remove location parameter from v1

oyamauchi:
## Description

This is the final step of migrating to separate top-level `zip` and
`address` parameters: removing the old nested parameter.

The embed frontend is already switched over, and there should be no
other clients of v1, so although this is a breaking change it should
be safe.

## Test Plan

`yarn lint`, `yarn test`. Run the update-fixtures script and make sure
the updates all succeed and there are no changes.

### DIFF
--- a/src/schemas/v1/utilities-endpoint.ts
+++ b/src/schemas/v1/utilities-endpoint.ts
@@ -1,7 +1,4 @@
-import {
-  API_REQUEST_LOCATION_SCHEMA,
-  API_RESPONSE_LOCATION_SCHEMA,
-} from './location';
+import { API_RESPONSE_LOCATION_SCHEMA } from './location';
 
 export const API_UTILITIES_RESPONSE_SCHEMA = {
   $id: 'APIUtilitiesResponse',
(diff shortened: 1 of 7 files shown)


veekas:
Looks like Oregon-specific tests are failing

src/schemas/v1/utilities-endpoint.ts
```diff
@@ -58,10 +53,10 @@ export const API_UTILITIES_SCHEMA = {
       },
     },
     oneOf: [
-      { required: ['location'] },
       { required: ['zip'] },
       { required: ['address'] },
     ],
+    additionalProperties: false,
```
veekas, line 59: I'm assuming this is just to have some additional security?

yarn.lock
```diff
@@ -1629,11 +1629,40 @@ doctrine@^3.0.0:
   dependencies:
     esutils "^2.0.2"
 
+dom-serializer@^2.0.0:
```
veekas, line 1632: oh nice, thanks for checking in this yarn lock update

veekas (approved):

oyamauchi:
BTW I checked the Cloud Run logs to make sure we completely stopped seeing this parameter when I merged rewiringamerica/embed.rewiringamerica.org#144

PR #370, opened 2024-03-18
Launch VA and DC to launched states

veekas:
Moves VA and DC to the list of launched states

Next step: after launch, remove "beta" state status in tests/fixtures, follow up on Spanish translations

### DIFF
--- a/src/data/types/states.ts
+++ b/src/data/types/states.ts
@@ -56,20 +56,20 @@ export const BETA_STATES: string[] = [
   'AZ',
   'CO',
   'CT',
-  'DC',
   'GA',
   'IL',
   'MI',
   'NV',
   'NY',
   'OR',
-  'VA',
   'VT',
   'WI',
 ];
 
 export const LAUNCHED_STATES: string[] = [
+  'DC',
   'RI',
+  'VA',
 ];
 
 export const isStateIncluded = (
--- a/test/data/schemas.test.ts
+++ b/test/data/schemas.test.ts
@@ -238,10 +238,6 @@ test("launched states do not have any values that we don't support for broader c
   STATE_INCENTIVE_TESTS.forEach(([state, , data]) => {
     if (LAUNCHED_STATES.includes(state)) {
       for (const incentive of data) {
-        // TODO: remove once City/County incentives have been beta-tested.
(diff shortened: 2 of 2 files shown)


veekas:
@datu925 looks like tests are failing because of [this test](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/cc428821943fdd514c6bacba90c477bcc64aa648/test/data/schemas.test.ts#L241-L243):

```ts
        // TODO: remove once City/County incentives have been beta-tested.
        tap.not(incentive.authority_type, AuthorityType.City);
        tap.not(incentive.authority_type, AuthorityType.County);
```

Is it okay to remove these checks now that VA/DC have been internally beta-tested?

datu925:
> @datu925 looks like tests are failing because of [this test](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/cc428821943fdd514c6bacba90c477bcc64aa648/test/data/schemas.test.ts#L241-L243):
> 
> ```ts
>         // TODO: remove once City/County incentives have been beta-tested.
>         tap.not(incentive.authority_type, AuthorityType.City);
>         tap.not(incentive.authority_type, AuthorityType.County);
> ```
> 
> Is it okay to remove these checks now that VA/DC have been internally beta-tested?
> 
> If it's helpful, here's the PR where this comment was added: https://github.com/rewiringamerica/api.rewiringamerica.org/pull/272/files#diff-1761174a5be653b788ac071e224ae0111ce6cbabd90dd3da3ddacf8583e7ec9dR202-R204

Yes, I think it's appropriate to remove those now. Thanks for checking.

oyamauchi:
My concern is we were going to add a disclaimer about possibly-inaccurate local incentive eligibility in the frontend, and we haven't done it yet (I'm not sure we even have a design for it?): https://app.asana.com/0/1206661332626418/1206661332626438

Unless we're close to external deadlines on these states, I'd prefer to have the disclaimer done before launching them. If we have deadlines to hit, then I'd say we can launch, but we should move the disclaimer to top priority in that case.

oyamauchi (approved):

oyamauchi:
Discussed offline: we're launching the states and prioritizing the disclaimer.

PR #371, opened 2024-03-19
Add Pennsylvania incentives

rpartridge:
Adds PA incentives. Spreadsheet: https://docs.google.com/spreadsheets/d/1H4hVUAwmEshbndeY2lFdnyk53qPZhoR5o1du5G1DVMw/edit#gid=616584455

Note this spreadsheet was initially generated by the LLM scripts but I manually reviewed and updated every incentive. Because this data is all new, it should all be up to date and ready to be released/tested by external partners as soon as this is in.

Ticket: https://app.asana.com/0/1206629812394927/1206778652392945/f


scripts/lib/spreadsheet-mappings.ts
```diff
@@ -132,6 +132,7 @@ export const VALUE_MAPPINGS: { [index: string]: AliasMap } = {
     ton: ['Ton'],
     watt: ['Watt'],
     btuh10k: ['Btuh10k'],
+    cmf50: ['CMF50'],
```
rpartridge, line 135: I added this because PA-34 has the incentive "$200 rebate for air sealing. Air infiltration reduction (@ CFM50) x $0.25 up to a max of $200" ([source](https://www.pplelectricsavings.com/ppl/sites/ppl/files/2023-08/2023_PPL_Res_ResReb_3862941_FLY_Q3_PRJ101498.pdf)). I looked this up and I think it's a unit?

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 500
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Up to $500 rebate for ENERGY STAR certified air source heat pumps meeting qualifications. Rebate value depends on efficiency."
```
oyamauchi, line 440: `maximum: 500` for this one too, I think

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 150
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$150 rebate for ENERGY STAR® certified packaged terminal heat pumps."
```
oyamauchi, line 514: I don't even know what this is? Since the item is `other` maybe we should skip it?

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+    "amount": {
+      "type": "percent",
+      "number": 1
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Free virtual home energy assessment. Assessment comes with a free energy savings kit mailed to home."
```
oyamauchi, line 656: I'm not sure if we include this kind of thing?

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+    }
+  },
+  {
+    "id": "PA-29",
+    "authority_type": "utility",
+    "authority": "pa-ppl-electric-utilities-corp",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "other",
```
oyamauchi, line 666: This is `heat_pump_air_conditioner_heater` I think? I forget where we stand on incentives for equipment that only cools and doesn't heat, but if we want this to show up, it should have that `item` value.

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+      "unit": "cmf50",
+      "maximum": 200
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "$200 rebate for air sealing. Air infiltration reduction (@ CFM50) x $0.25 up to a max of $200."
+    },
```
oyamauchi, line 794: Now that I'm looking at it, this is going to be nonsense to a layperson. It will show in the frontend as "$0.25/CFM50 up to $200 for weatherization".

Maybe we drop the per-unit part, code it as `maximum: 200` and only talk about CFM50 in the description?

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+    "start_date": "2024",
+    "end_date": "2024"
+  },
+  {
+    "id": "PA-42",
+    "authority_type": "utility",
+    "authority": "pa-peco",
+    "payment_methods": [
+      "pos_rebate"
```
oyamauchi, line 959: Is this really an upfront discount? If not it should be `rebate`. (Same question for the few other `pos_rebate` incentives.)

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 300
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Up to $300 rebate for air source heat pumps. Rebate value depends on the equipment efficiency."
```
oyamauchi, line 995: `maximum: 300` I think (same for PA-44)

data/PA/incentives.json
```diff
@@ -0,0 +1,1023 @@
...
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Up to $300 rebate for central air conditioner. Rebate value depends on equipment efficiency."
+    },
+    "bonus_available": true
+  },
+  {
+    "id": "PA-30",
```
oyamauchi, line 682: This isn't something to solve in this PR, but this is not ideal. PA-30 through PA-33 will show up in the frontend as four cards:

- "75% up to $200 for weatherization"
- "75% up to $500 for weatherization"
- "75% up to $200 for weatherization"
- "75% up to $500 for weatherization"

...all with the same program name.

It will look terrible and the actual distinctions (basement vs. attic, electric vs. non-electric heat) will only be in the fine print.

I thought one possible quick fix would be to make the electric vs. non-electric heat pairs mutually exclusive, since your house is either electric-heated or not. But that would be equivalent to completely omitting one of them, since the calculation logic would find either both or neither ineligible on the basis of other factors. (I think we talked about this problem on one of your other PRs?)

Really don't know what to do here! I've been thinking about adding a concept of sub-item (which would allow a distinction between attic and basement insulation), but that doesn't help answer what this should look like in the frontend.

oyamauchi (changes requested):
It looks like you need to run `scripts/generate-utility-data.ts` -- the list of utilities in authorities.json is a subset of what's in the sqlite DB, and they're supposed to be the same now. You'll need to do a bit of cleanup -- e.g. in the DB I see both `pa-peco` and `pa-peco-energy`, which I bet are the same company -- as documented in scripts/README.md. (This is necessary so that the zip-to-utility mapping works right.)

data/PA/incentives.json
rpartridge: this has come up a couple times in QA sessions as well and we haven't been totally sure how to handle it. I think we have talked about making electric vs non electric mutually exclusive, but I agree that isn't the right call here (although if we coded these differently in the API we could show different incentives in PEP where we do know a user's heating source).

I think the easiest fix right now is to combine these 4 into 2 incentives, 1 for electric and 1 for non electric, and add more details to the description. something like:
- PA-30: Rebate for 75% of costs up to $1000 for insulation in an electric heated house. Up to $500 for attic insulation and $500 for basement wall insulation.
- PA-31: Rebate for 75% of costs up to $400 for insulation (central AC, non-electric heat). Up to $200 for attic insulation and $200 for basement wall insulation.
- Delete PA-32 and PA-33

thoughts?

oyamauchi: Yeah, I like that. I think the description should start with the electric/non-electric distinction and have the attic/basement part after.

oyamauchi (approved):

PR #373, opened 2024-03-20
Add way to read Google Sheets API and save hyperlinks into JSON.

datu925:
This is half of the Google Sheets <-> JSON conversion. The other half is coming shortly.

Welcoming @mmezher's thoughts but @oyamauchi, I think it's best if you're the primary reviewer. Since you'll be absorbing any spreadsheet-json work, I want to make sure you're on board.

As mentioned in [this thread](https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1710441552470399), hyperlinks are a problem for the conversion since they are not easily stored in JSON. This is especially true because you can have multiple links in the same cell, and it becomes difficult to store this without making it very difficult for humans to edit the JSON. So as decided in that thread, we're condensing the hyperlinks to text.

To actually deploy this new method, I've changed the spreadsheet-to-json flow so that we start saving the CollectedIncentives (spreadsheet equivalents) for states that specify an additional collectedFilepath in the spreadsheet registry. There's also some Google API setup. I haven't documented this much in the README; I'm hoping to get the second half in place before making this widely available.

### DIFF
--- a/scripts/incentive-spreadsheet-to-json.ts
+++ b/scripts/incentive-spreadsheet-to-json.ts
@@ -1,6 +1,7 @@
 import Ajv from 'ajv';
 import { parse } from 'csv-parse/sync';
 import fs from 'fs';
+import { google, sheets_v4 } from 'googleapis';
 import fetch from 'make-fetch-happen';
 import minimist from 'minimist';
 import path from 'path';
@@ -17,10 +18,13 @@ import {
 } from '../src/data/state_incentives';
 import { LOCALIZABLE_STRING_SCHEMA } from '../src/data/types/localizable-string';
 import { FILES, IncentiveFile } from './incentive-spreadsheet-registry';
+import { authorize } from './lib/auth-helper';
 import { DataRefiner } from './lib/data-refiner';
 import {
   CollectedIncentivesWithErrors,
+  LinkMode,
   flatToNestedValidate,
+  googleSheetToFlatData,
 } from './lib/format-converter';
 import { FIELD_MAPPINGS, VALUE_MAPPINGS } from './lib/spreadsheet-mappings';
 import { SpreadsheetStandardizer } from './lib/spreadsheet-standardizer';
@@ -34,6 +38,7 @@ export type StateIncentivesWithErrors = Partial<StateIncentive> & {
 
 type SpreadsheetConversionOutput = {
   invalidCollectedIncentives: CollectedIncentivesWithErrors[];
+  validCollectedIncentives: CollectedIncentive[];
   invalidStateIncentives: StateIncentivesWithErrors[];
   validStateIncentives: StateIncentive[];
 };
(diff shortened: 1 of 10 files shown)


scripts/lib/format-converter.ts
```diff
@@ -122,3 +124,103 @@ function coerceToBoolean(input: string): boolean | string {
...
+  const cellAddressToHyperlinks: { [index: string]: string[] } = {};
+  const records: Record<string, string>[] = [];
+  const headers: string[] = [];
+  incentives.data[0].rowData!.forEach((row, rIndex) => {
+    if (rIndex < headerRow - 1) return; // skip pre-header rows
+    if (rIndex === headerRow - 1) {
+      if (!row.values)
+        throw new Error(
+          'No values found in header row, probably due to incorrect header row provided',
+        );
```
oyamauchi, line 184: I think we've pretty consistently avoided brace-less control flow statements; that's certainly my preference. We can add this as an eslint rule and fix it all automatically, so you don't need to fix it here; I'll follow up with the eslint change.

scripts/incentive-spreadsheet-to-json.ts
```diff
@@ -127,15 +176,29 @@ async function convertToJson(
...
-  const response = await fetch(file.sheetUrl);
-  const csvContent = await response.text();
-  const rows = parse(csvContent, {
-    columns: true,
-    from_line: file.headerRowNumber ?? 1,
-  });
+
+  // any needed given output of csv-parse
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let rows: any;
```
oyamauchi, line 182: Couldn't it be `Record<string, string>[]`? That's the type of the `spreadsheetToJson` param that `rows` gets passed to.

scripts/lib/auth-helper.ts
```diff
@@ -0,0 +1,66 @@
+import { authenticate } from '@google-cloud/local-auth';
+import fs from 'fs';
+import { AuthClient, OAuth2Client } from 'google-auth-library';
+import { google } from 'googleapis';
+
+const TOKEN_PATH = 'secrets/token.json';
+const CREDENTIALS_PATH = 'secrets/credentials.json';
```
oyamauchi, line 7: Might be worth doing these as

```ts
const TOKEN_PATH = path.join(__dirname, '../../secrets/token.json');
```

Just to be robust against the script being run from an unexpected working directory.

oyamauchi (approved):
I'm not at all familiar with the Google auth stuff, so I'm just trusting that that makes sense.

Otherwise this seems good. I'll look at the other half tomorrow.

scripts/incentive-spreadsheet-to-json.ts
```diff
@@ -110,11 +115,55 @@ export function spreadsheetToJson(
...
 }
 
+type UrlParts = {
+  spreadsheetId: string;
+  incentiveDataSheetId: number;
+};
+
+const urlRegex = new RegExp(
+  /https:\/\/docs\.google\.com\/spreadsheets\/d\/(.*)\/pub\?gid=([0-9]+)&.*/,
+);
```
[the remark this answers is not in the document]
oyamauchi: Well, one reason to use the RegExp constructor is to pass it a normal string as the pattern; that way you don't have to escape all the forward slashes. You could do that here since there aren't any characters in the regex that would have special meaning in a normal quoted string.

test/scripts/incentive-spreadsheet-to-json.test.ts
```diff
@@ -127,3 +130,13 @@ test('drop records with omit_from_api marked', tap => {
   tap.equal(validStateIncentives.length, 0);
   tap.end();
 });
+
+test('extract IDs from a standard URL', tap => {
+  const url =
```
mmezher, line 135: add test for failure cases, of failed spreadsheet ID, incentive data sheet ID, or not matching docs regex?

Probably best to cover all bases here.

scripts/incentive-spreadsheet-to-json.ts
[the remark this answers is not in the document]
datu925: Done. I had gotten tripped up with the string version because I forgot you had to `\\` the question mark that starts the query params.

test/scripts/incentive-spreadsheet-to-json.test.ts
datu925: Done. I had to reconfigure things and introduce an interface so I could fake the sheets client.

mmezher: Thanks! Tests LGTM.

mmezher (approved):
LGTM for feedback addressed.

PR #374, opened 2024-03-20
Add method to export collected incentives to a Google sheet.

datu925:
This is the other half of Google Sheet <-> JSON conversion; a continuation of https://github.com/rewiringamerica/api.rewiringamerica.org/pull/373. They're largely independent, though they both rely on the authentication helper and the changes to the spreadsheet-registry.

This is the CollectedIncentives -> Google Sheet direction. It uses the Google Sheets API to create a new spreadsheet. Right now, we only have the incentives data itself, but as a follow-up, I'll create other worksheets like the data model and the standardized enum values so that it is more comparable to our existing sheets (plus add data validation).

Notable assumption: the first alias in the FIELD_MAPPINGS and VALUE_MAPPINGS in spreadsheet-mappings.ts is considered the canonical human-readable spreadsheet version. I think this is fine for now, though we may want clearer documentation of it at some point.

Testing: unit tests, and I tried this on Colorado. You can see the round-trip in [this spreadsheet](https://docs.google.com/spreadsheets/d/1vybxSsCChFp1iAwMy1CwUFEjAxh0KxAmoDl7grtlKQ8/edit#gid=406119121). This was running the Sheets -> CollectedIncentives version followed by CollectedIncentives -> Sheets, then copying over the original CO data and comparing. The green cells are where there's a difference (the new value is on top).

 The cell values are the same, except in:
1. places where the original Colorado data used a different alias from the canonical alias
2. dates. The original sheet has these as numbers formatted as dates; the new version has them as strings. In keeping with Owen's direction for dates, I think we want them to be strings anyway.
3. columns missing from the Colorado spreadsheet - there are some, like Editorial Notes, that aren't in the original CO data, but in the interest of standardizing our spreadsheets, I think it's good to be adding them back in.
4. Cells that had hyperlinks (as expected)

### DIFF
--- a/scripts/lib/format-converter.ts
+++ b/scripts/lib/format-converter.ts
@@ -237,24 +237,54 @@ export function googleSheetToFlatData(
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type StringKeyed = { [index: string]: any };
 
+// Reverse of flatToNested methods above. Note that arrays are not expanded, so
+// if we add arrays of nested objects to the schema, we will need to revisit
+// this.
 export function nestedToFlat(input: StringKeyed): StringKeyed {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const output: { [index: string]: any } = {};
-  for (const [fieldName, field] of Object.entries(input)) {
-    if (field === undefined) continue;
-    if (Array.isArray(field)) {
-      output[fieldName] = field.join(', ');
-    } else if (typeof field === 'object' && !Array.isArray(field)) {
+  const output: StringKeyed = {};
+  for (const [fieldName, fieldValue] of Object.entries(input)) {
+    if (fieldValue === undefined) continue;
+    if (typeof fieldValue === 'object' && !Array.isArray(fieldValue)) {
       for (const [nestedKey, nestedVal] of Object.entries(
-        nestedToFlat(field),
+        nestedToFlat(fieldValue),
       )) {
         output[`${fieldName}.${nestedKey}`] = nestedVal;
       }
     } else {
-      output[fieldName] = field;
+      output[fieldName] = fieldValue;
     }
   }
   return output;
+}
+
+function createCellValue(
+  colName: string,
+  colValue: object | string | number | boolean,
+): sheets_v4.Schema$ExtendedValue {
+  let valToWrite: sheets_v4.Schema$ExtendedValue = {};
+  if (Array.isArray(colValue)) {
+    valToWrite = { stringValue: colValue.join(', ') };
+  } else if (typeof colValue === 'string') {
+    valToWrite = { stringValue: colValue };
+  } else if (typeof colValue === 'boolean') {
+    valToWrite = { boolValue: colValue };
+  } else if (typeof colValue === 'number') {
+    valToWrite = { numberValue: +colValue };
+  } else if (typeof colValue === 'object') {
+    throw new Error(
+      `Trying to write object to spreadsheet. Objects should be flattened prior to write. Column name: ${colName}; value: ${util.inspect(
+        colValue,
+      )}`,
+    );
+  } else {
+    console.warn(
+      `Unexpected value type: ${colName}: ${util.inspect(
+        colValue,
+      )}. Writing as string.`,
+    );
+    valToWrite = { stringValue: colValue };
+  }
+  return valToWrite;
 }
 
 export function collectedIncentiveToGoogleSheet(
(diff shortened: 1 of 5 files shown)


scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -33,10 +36,12 @@ export class SpreadsheetStandardizer {
     valueMap: { [index: string]: AliasMap } = VALUE_MAPPINGS,
     strict: boolean = true,
   ) {
+    this.fieldMap = fieldMap;
```
mmezher, line 39: I think it might be worth adding a comment about the uses of the field/val maps, and their reversed maps. AFAICT the original maps are now saved due to wanting to convert back to aliases? but since the original names were swapped out, it took a bit.

scripts/lib/format-converter.ts
```diff
@@ -224,3 +230,93 @@ export function googleSheetToFlatData(
   }
   return records;
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type StringKeyed = { [index: string]: any };
+
+export function nestedToFlat(input: StringKeyed): StringKeyed {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const output: { [index: string]: any } = {};
```
mmezher, line 239: value? or fieldValue?

datu925: Sorry, I don't understand what this comment is referring to.

scripts/lib/spreadsheet-standardizer.ts
datu925: Fair, this was probably a pretty awkward diff given the original names weren't great. Added some comments above the constructor.

scripts/lib/format-converter.ts
mmezher: oh, sorry, I probably saved it to the wrong line.

for (const [fieldName, field]) 

this is referring to renaming 'field'. Specifically due to believing that if fieldName exists (since it's in input's entries) it seems weird to check if the field itself is undefined (when really it's checking if the value of the field doesn't exist here, and is checking the contents of the named field).


Does that make sense?

scripts/lib/spreadsheet-standardizer.ts
mmezher: Does this need to refresh? I don't see the comments.

mmezher: Oops, I was on the former commit. LGTM!

scripts/lib/format-converter.ts
```diff
@@ -224,3 +230,93 @@ export function googleSheetToFlatData(
...
+    false,
+  );
+
+  const headers = Object.values(fieldMappings).map(mapping => mapping[0]);
+  let rowData: sheets_v4.Schema$RowData[] = [];
+  incentives.forEach(incentive => {
+    const flattened = nestedToFlat(incentive);
+    const aliased = standardizer.convertToAliases(flattened);
+    // First ensure we have a header column for every value, including those
+    // not in our schema.
```
oyamauchi, line 273: How are we going to handle the fact that the existing spreadsheets generally have the headers on row 2? Reading these generated spreadsheets back would require a different headerRowNumber.

scripts/lib/spreadsheet-standardizer.ts
```diff
@@ -79,17 +89,33 @@ export class SpreadsheetStandardizer {
...
       output[newCol] = this.handleSpecialCases(val, newCol);
     }
     return output;
   }
+
+  convertToAliases(input: StringKeyed): StringKeyed {
+    const output: StringKeyed = {};
+    for (const [fieldName, field] of Object.entries(input)) {
+      let outputFieldName = fieldName;
+      if (this.fieldMap[fieldName] && this.fieldMap[fieldName][0]) {
```
oyamauchi, line 108: Wouldn't it be an error condition if this were false? It would mean there's a field in the JSON that isn't mapped to/from a spreadsheet column, and that shouldn't happen, right?

scripts/lib/format-converter.ts
```diff
@@ -224,3 +230,93 @@ export function googleSheetToFlatData(
...
+    for (const key in aliased) {
+      if (!headers.includes(key)) {
```
oyamauchi, line 275: This seemed weird to me -- it seemed like it shouldn't be possible -- which is what led me to the comment in `convertToAliased`. I might not have understood all of what's going on here.

scripts/lib/format-converter.ts
```diff
@@ -224,3 +230,93 @@ export function googleSheetToFlatData(
...
+          valToWrite = { stringValue: aliased[col] };
+        } else if (typeof aliased[col] === 'boolean') {
+          valToWrite = { boolValue: aliased[col] };
+        } else if (typeof aliased[col] === 'number') {
+          valToWrite = { numberValue: +aliased[col] };
+        } else {
+          console.warn(
+            `Unexpected value type: ${col}: ${aliased[col]}. Writing as string.`,
+          );
+          valToWrite = { stringValue: aliased[col] };
```
oyamauchi, line 296: You might want a `util.inspect` call here, or something like that? What happens if `aliased[col]` is an object or array; what shows up in the spreadsheet?

oyamauchi (commented):
This PR doesn't include the coloring of the cells, right? Did you do that manually in the demonstration sheet?

Not requesting changes because there's nothing I think _needs_ to change before landing, but I'd like to understand the commented parts a little better before approving.

scripts/lib/spreadsheet-standardizer.ts
datu925: See my response to your related comment – for collected data, we unfortunately do expect there may be columns that aren't part of our schema, and I think we should support that (Tom C told me as much).

scripts/lib/format-converter.ts
oyamauchi: Ah, OK.

It occurs to me, though, that the word/character count columns will be troublesome because they're formulas. If the sheet-to-JSON process just captures the displayed text, then JSON-to-sheet will render them as "hardcoded" values that don't change when the description does. On the whole I don't think it's the worst problem, but something to be aware of.

oyamauchi (approved):
OK, I'm good to proceed with this.

datu925:
That's a good point re: the formulas. There are some potential follow-ups since they're easy to detect, so we can at least warn the user that they're being dropped and/or converted to values.

With extra engineering effort, lots of these sorts of things can survive the spreadsheet-json border, but in general I think it's best to not support them unless they turn out to be must-haves.

PR #375, opened 2024-03-20
Don't use the fs/promises API in generate-utility-data

oyamauchi:
## Description

There's a filesystem error upon reading the authorities.json file
towards the end of the script. (I don't know why I didn't see this
during testing?) Switching to the synchronous fs APIs seems to fix it,
so just do that rather than sink more time into investigating. None of
these reads/writes would happen concurrently anyway, so there's no
real downside to this.

## Test Plan

Run the script and don't get an `EBADF` error.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -34,7 +34,17 @@ Filling out an entry for `incentive-spreadsheet-registry.ts` consists of creatin
 - Optionally declaring the header row number, if not the top row of the spreadsheet, in `headerRowNumber`
 
 First, **edit `authorities.json` manually** to include a top-level entry for
-the state you're adding, if it's not already present.
+the state you're adding, if it's not already present, like this:
+
+```json
+{
+  "<state-abbreviation>": {
+    "state": {},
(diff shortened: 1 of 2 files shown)


rpartridge (approved):
fixed the file system error! thanks for the quick fix. approving because code lgtm but a couple suggestions:

1. could you add to the readme to clarify that in order to get this to run, you need to add:
```
"<state>": {
    "state": {},
    "utility": {}
  },
```
before running the script? based on the existing [docs](https://github.com/rewiringamerica/api.rewiringamerica.org/tree/main/scripts#spreadsheet-to-json) it looks like you just need to add `"<state>": {}` which was failing for me

2. running the script locally modified authorities.json and added a bunch of utilities in Oregon. should we make a ticket to update OR to include these?

oyamauchi:
> running the script locally modified authorities.json and added a bunch of utilities in Oregon. should we make a ticket to update OR to include these?

Yeah, we should land that as a separate PR.

PR #377, opened 2024-03-20
Consolidate NV utilities

rpartridge:
I was doing 1 final test before emailing NCEF about testing when I noticed that a lot of zip codes were popping up with utilities I'd never heard of (Nevada Power, Sierra Pacific Power). I did some quick googling and noticed that those utilities were consolidated into NV Energy in 2008 and as far as I can tell no longer exist as separate entities. This PR collapses Nevada Power and Sierra Pacific Power into NV Energy so that customers in these zip codes see the NV Energy incentives

Sources:
- [NV energy website](https://www.nvenergy.com/about-nvenergy/our-company/history)
- [Wikipedia](https://en.wikipedia.org/wiki/Nevada_Power_Company)


data/authorities.json
```diff
@@ -829,9 +829,6 @@
       "nv-mt-wheeler-power": {
         "name": "Mt Wheeler Power"
       },
-      "nv-nevada-power": {
```
rpartridge, line 832: Sierra Pacific Power didn't get removed because there's still 1 row in the spreadsheet that references `Sierra Pacific Power Co [Berkshire Hathaway Inc]` in NV with a NV zip code. I don't think we can override this one because unfortunately that utility also exists in California with the same name and utility code, and the CA utility is actually legitimate:
Screenshot 2024-03-20 at 1 23 26 PM

functionally this shouldn't matter since it only applies to 1 zip code, 89803, and the primary utility for that zip code is Mt Wheeler power so it wouldn't show up in the dropdown anyway. it's just a little bit weird that Sierra Pacific still shows up in authorities.json

oyamauchi: That kinda feels to me like an error in the dataset. Sometimes you see cases like this when the ZIP is right on a state border (e.g. a little piece of VA on the KY border is served by Kentucky Utilities), but 89803 isn't anywhere near the CA/NV border. I agree it's OK to let this slide.

oyamauchi (approved):
Nice. In general I think the most recent round of states have been done without vetting the utility data; that might be something we ought to have the PEP engineers go back and do before they roll off.

rpartridge:
> In general I think the most recent round of states have been done without vetting the utility data; that might be something we ought to have the PEP engineers go back and do before they roll off.

agreed! I can make some tickets on our board https://app.asana.com/0/1206629812394927/1206659604468667

PR #378, opened 2024-03-21
Add <p> tag check to return relevant contents.

mmezher:
<p> tag in HTML incentive sites generally have the relevant information, including the last updated time. From checking numerous websites, I believe this will be sufficient for figuring out whether content has changed or should be re-reviewed more generally (with the opportunity to add specific matchers for different websites later on).


scripts/lib/website-comparisons.ts
```diff
@@ -6,6 +6,10 @@ export function cleanWebsiteData(response: string): string {
   const $ = cheerio.load(response);
   // Only check for the HTML body to try and avoid metadata and scripts that may change per-system/call.
   if ($('body').length > 0) {
+    // If <p> tags exist, try to only return those, as they'll likely be relevant tags.
+    if ($('p').length > 0) {
+      return $('p').contents().text().trim();
```
datu925, line 11: I don't have an intuition about what this looks like if there are multiple <p> tags. One option is for me to just not worry about it as long as it works, since this text is hashed anyway. But if you think there is a comment that might succinctly describe what the output of this line is (is it concatenated together? Is it an array?), that might help.

datu925 (approved):

scripts/lib/website-comparisons.ts
mmezher: There sure is! Added comment.

PR #379, opened 2024-03-21
Add OR utilities using script

rpartridge:
This adds the OR utilities using the new `generate-utility-data` script. These weren't added in https://github.com/rewiringamerica/api.rewiringamerica.org/pull/368 , which added all utilities for all other existing states (I'm guessing just a timing issue since the OR incentives got merged around then). When I was running this script for other states, it always adds these OR utilities to the diff since they're missing. We should just get this in now so that when we use the script in the future the only diffs are for new states we're adding

## Considerations
1. This shouldn't change existing functionality, the 2 OR utilities that were added before are unchanged
2. I glanced briefly through the new utilities and don't think any of them need to be overridden/modified


data/authorities.json
```diff
@@ -1034,11 +1034,125 @@
...
+        "name": "City of Milton-Freewater"
+      },
+      "or-city-of-monmouth": {
+        "name": "City of Monmouth"
+      },
+      "or-city-of-springfield": {
+        "name": "City of Springfield"
+      },
+      "or-clatskanie-peoples-util-dist": {
+        "name": "Clatskanie Peoples Util Dist"
```
oyamauchi, line 1080: "Clatskanie PUD": https://www.clatskaniepud.com/

data/authorities.json
```diff
@@ -1034,11 +1034,125 @@
...
+        "name": "Consumers Power"
+      },
+      "or-coos-curry-electric-cooperative": {
+        "name": "Coos-Curry Electric Cooperative"
+      },
+      "or-douglas-electric-cooperative": {
+        "name": "Douglas Electric Cooperative"
+      },
+      "or-emerald-people-s-utility-dist": {
+        "name": "Emerald People's Utility Dist"
```
oyamauchi, line 1107: Spell out "District"

data/authorities.json
```diff
@@ -1034,11 +1034,125 @@
...
+        "name": "Portland General Electric"
+      },
+      "or-salem-electric": {
+        "name": "Salem Electric"
+      },
+      "or-surprise-valley-electrification": {
+        "name": "Surprise Valley Electrification"
+      },
+      "or-tillamook-peoples-utility-dist": {
+        "name": "Tillamook Peoples Utility Dist"
```
oyamauchi, line 1149: Spell out "District": https://www.tpud.org/

data/authorities.json
```diff
@@ -1034,11 +1034,125 @@
...
+        "name": "Midstate Electric Cooperative"
+      },
+      "or-northern-wasco-county-pud": {
+        "name": "Northern Wasco County PUD"
+      },
+      "or-nw-natural": {
+        "name": "NW Natural"
+      },
+      "or-oregon-trail-el-cons-cooperative": {
+        "name": "Oregon Trail El Cons Cooperative"
```
oyamauchi, line 1134: "Oregon Trail Electric Cooperative": https://otec.coop/

data/authorities.json
```diff
@@ -1034,11 +1034,125 @@
...
+        "name": "Canby Utility Board"
+      },
+      "or-central-electric-cooperative": {
+        "name": "Central Electric Cooperative"
+      },
+      "or-central-lincoln-people-s-ut-dt": {
+        "name": "Central Lincoln People's Ut Dt"
+      },
+      "or-central-lincoln-peoples-utility-district": {
+        "name": "Central Lincoln Peoples Utility District"
```
oyamauchi, line 1050: This is the same as the one above

data/authorities.json
```diff
@@ -1034,11 +1034,125 @@
       }
     },
     "utility": {
-      "or-portland-general-electric": {
-        "name": "Portland General Electric"
+      "or-blachly-lane-county-cooperative-el-association": {
+        "name": "Blachly-Lane County Cooperative El Association"
```
oyamauchi, line 1038: "Blachly-Lane Electric Cooperative": https://www.blachlylane.coop/

oyamauchi (changes requested):
We should at least clean up the weird abbreviations; I googled those and found their customer-facing names. Also found two that are the same

oyamauchi (approved):

PR #380, opened 2024-03-21
Add CA incentives

matelau:
Spreadsheet is here: https://docs.google.com/spreadsheets/d/1i7xKnAbZ8QyXQkGSHeK3qIYTY2J3HgHyoT12GVcDZao/edit?usp=sharing

Ticket: https://app.asana.com/0/1206629812394927/1206662027052986/f


data/CA/incentives.json
```diff
@@ -0,0 +1,2858 @@
...
+    "payment_methods": [
+      "assistance_program"
+    ],
+    "item": "weatherization",
+    "program": "ca_bearValleyElectricService_energySavingsAssistanceProgram(ESAP)",
+    "amount": {
+      "type": "percent",
+      "number": 100
+    },
+    "owner_status": [
```
matelau, line 15: I may need to leverage this mapping for homeowner vs renter, but this may require going through each authority and figuring out if they explicitly state homeowner vs tenant  https://github.com/rewiringamerica/incentive-data-llm/blob/main/src/constants.ts#L16

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+      "type": "dollar_amount",
+      "number": 500,
+      "minimum": 500,
+      "maximum": 1000
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "$500 rebate for the installation of an electric heat pump water heater if replacing an electric resistance water heater."
```
oyamauchi, line 2398: Why might this go up to $1000?

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+      "type": "dollars_per_unit",
+      "number": 3.75,
+      "unit": "kilowatt",
+      "maximum": 11250
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Income-Qualified Solar Grant Program provides a grant of $3.75 per watt, up to 3 kilowatts DC."
```
oyamauchi, line 2440: Unit confusion here. The `unit` should be `watt` if that's how the utility is presenting it. If they're presenting it as kW, then the unit should be `kilowatt` and the number should be 3750.

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "electric_panel",
+    "program": "ca_siliconValleyPower_rebates",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 2000,
+      "minimum": 2000,
+      "maximum": 4000
```
oyamauchi, line 2457: I'd remove the max here and, if you can claim this twice, note that in the description. (Same deal as with the smart thermostat incentives.)

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "program": "ca_bayAreaRegionalEnergyNetwork(BayREN)_rebates:SingleFamilyHomeowners",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 500
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "$500 rebate for duct replacement."
```
oyamauchi, line 2602: Ooh, this is a new one. Good to leave it as "other" for now, but we'll probably want a new `item` value for this.

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+  },
+  {
+    "id": "CA-124",
+    "authority_type": "other",
+    "authority": "ca-bay-area-regional-energy-network-bay-ren",
+    "eligible_geo_group": "ca-group-bay-area-regional-energy-network-bay-ren",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "other",
```
oyamauchi, line 2614: Should be `heat_pump_air_conditioner_heater`, right?

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "amount": {
+      "type": "dollars_per_unit",
+      "number": 0.2,
+      "unit": "square_foot"
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Attic Insulation upto 20¢ per sq. ft. Minimum Rating R-19."
```
oyamauchi, line 2669: "up to", and let's go with $0.20 to be consistent. (I'd actually consider switching all dollar amounts less than 1 to be expressed as ¢, but I'd rather do that as an across-the-board change and not have this as a one-off.)

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "amount": {
+      "type": "dollars_per_unit",
+      "number": 0.15,
+      "unit": "square_foot"
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Exterior Wall Insulation upto 20¢ per sq. ft. Minimum Rating R-13+/."
```
oyamauchi, line 2690: Numbers don't match

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "amount": {
+      "type": "dollar_amount",
+      "number": 1000
+    },
+    "owner_status": [
+      "homeowner",
+      "renter"
+    ],
+    "short_description": {
+      "en": "Up to $1,000 rebate for a used Battery Electric or Plug-In Hybrid vehicle if purchased or leased after January 1, 2023."
```
oyamauchi, line 2752: I don't think the "after January 1" part is necessary; that's long past

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "program": "ca_riversidePublicUtilities_residentialRebates",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 805
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Up to $805 rebate for installing a Time-of-Use (TOU) EV Rate Meter."
```
oyamauchi, line 2793: Another tricky one. Good to leave as `other` for now.

data/CA/incentives.json
```diff
@@ -0,0 +1,2817 @@
...
+    "amount": {
+      "type": "dollars_per_unit",
+      "number": 0.75,
+      "unit": "square_foot"
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Rebate for qualified new attic insulation upgrades."
```
oyamauchi, line 2814: What's qualified?

oyamauchi (changes requested):
Bunch of comments in the incentive data. Before addressing them individually, I'd recommend adding the new incentives file to `schemas.test.ts`; it will catch some of the things I flagged and possibly others.

.gitignore
```diff
@@ -8,4 +8,5 @@ yarn-error.log
 scripts/geodata
 build
 .tap
+.tool-versions
```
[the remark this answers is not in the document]
matelau: I've been using ASDF for language version management. So, I thought I would slide this in our gitignore to not accidentally check that in.

PR #383, opened 2024-03-22
Scripts and data for calculating AMI and 30c eligibility

mikivee:
This set of files is a self-contained portion of the repo that is used to create AMI/30C eligibility lookup tables. Follow the readme to download all necessary files into `data/raw`. See [this page](https://www.notion.so/rewiringamerica/AMI-MFI-for-Incentives-a18db19ff12840319472bc2803a63b31?pvs=4) for details on algorithm and terminology. 

There are two main scripts in this PR, both of which create tables that are written to `data/raw`:
* `ami.py` : creates AMI income threshold lookup tables at various levels of geographic aggregation. 
* `30c_eligibility.py`:   creates lookup table for 30c (EV charger tax credit) eligibility at zip and tract level. 

Outstanding questions for @oyamauchi : 
- Future-proofing code: Should downloaded files in `raw` be named so that they don't change in subsequent refreshes? What about column names within those downloaded files? (e.g, column named `ami2024`). 
- Should I add data checks for future data refreshes?: usually I would do assert statements.. is that okay? 
- Added some basic install instructions in the readme-- is this sufficient? 
- I unfortunately wasted a bit of time calculating the 30c tract eligibility from scratch, and got about 90% through coding it before I realized I could just download a CSV from Argonne. Since these tracts will be valid thru 2029, there won't be a need to refresh this. Since I already wrote most of it, should I check in the more complicated 30c code in case we ever need to calculate from scratch? 
- I decided that trying to do zips instead of ZCTAs was too complicated given issue with the USPS API. I think we can return to this eventually if you think the the benefit of increased spatial granularity warrants more putting in more cycles. 

[Link](https://docs.google.com/document/d/1VEtI4F6mYHy0p1mq9e4Pn0PWvQKWAE8vYMN9CM7OLGo/edit#heading=h.krkv38f9bzgc) to original request from Owen.


oyamauchi (changes requested):
On top-level questions:

* See inline comments re: naming files with the year. The overall gist is I'd rather people not have to go in there and edit a few specific lines of code next year. There are a couple of ways to accomplish that, which I'll leave up to you.
* I'm not sure exactly what the data checks would be, but if it's not too much trouble, sounds like a good idea. Assert statements are fine.
* I think the README is basically fine, modulo a few inline comments. If you end up adding the year as an argument to `ami.py`, put that in the README too.
* I'm on the fence about checking in the from-scratch 30C code. But if we don't expect to run it at all for the next 5 years, then I'm inclined to not bother. Who knows how much things will have changed (both in the codebase and in IRL policy) by 2029; I would bet it would need nontrivial updates by then anyway.
* Re: zip and zcta. At this point I think (a) the priority is to get something out, and (b) we're using zcta in production already, so I think this is fine, and we can consider switching to use zip a future enhancement (which, realistically, I don't know when we'll have time for).

scripts/income_limits/30c_eligibility.py
```diff
@@ -0,0 +1,76 @@
...
+# aggregate over tracts in each zip minimizing type II error"
+# zip is_eligible iff all intersecting tracts are eligible
+eligibility_30c_by_zcta = util.aggregate_over_origin(
+    df=eligibility_30c_by_tract_zcta,
+    groupby_cols=['zcta'],
+    agg_cols=['is_eligible'],
+    minimize_type_2_error=True)
+
+
+# -- 4. Write out data -- #
```
oyamauchi, line 66: Oh, can you have it write the boolean columns as `1` for true and `0` for false? These are going into SQLite, which doesn't have a boolean type, so a numeric true/false thing works better

scripts/income_limits/ami.py
```diff
@@ -0,0 +1,192 @@
...
+        dtype={col: str for col in string_cols}))
+    .rename({'zip_census_tabulation_area': 'zcta'}, axis=1)
+    .drop('county_name', axis=1)
+)
+
+# remove non-zctas
+zcta_countysub_crosswalk = zcta_countysub_crosswalk[
+    zcta_countysub_crosswalk.zip_code_name != '[not in a ZCTA]']
+
+# clean up geoids
```
tsgra, line 90: Suggestion: Normally I wouldn't ask for this, but since the formatting of the geoids is the whole ballgame here as far as mapping AMIs to the right geography, it might be helpful to have some more documentation as to how exactly you're formatting the geoids. It's not immediately apparent based on the code. 

Alternatively, it's more easily groccable for me if you put `zcta_countysub_crosswalk['county_code'] + zcta_countysub_crosswalk['county_subdivision_2020']`, but you may have performance reasons for structuring the code as you did.

tsgra (approved):
This is awesome! Way less brittle than my original methodology and more easily updated going forward. I just had a few suggestions around documentation and a suggestion on the `util.aggregate_over_origin` function, but none were blocking so approving!

scripts/income_limits/ami.py
```diff
@@ -0,0 +1,192 @@
...
+    ami_by_countysub[~is_county_msk]
+    .merge(
+        tract_countysub_crosswalk,
+        on='countysub_geoid')
+)
+ami_countysub_tract = pd.concat(
+    [ami_county_tract_non_new_england, ami_countysub_tract_new_england])
+# %%
+# Aggregate over countysubs to get AMI at target geography using strategy of minimizng type II error
+ami_by_zcta = util.aggregate_over_origin(
```
tsgra, line 145: Blocking: I attempted to use the other implementation - the weighted average over origin geographies, and it threw the following error: 

`TypeError: aggregate_over_origin.<locals>.weighted_average() got an unexpected keyword argument 'include_groups'` 

Seems you might not need the `include_groups` arg in the weighted average function?

Additionally, have you characterized the difference between the result using both methodologies? I attempted to here but wasn't able to resolve the error, but might be good to examine if adjacent counties or countysubs have wildly different AMI thresholds. I feel like it's unlikely, but would be good to know generally.

tsgra (changes requested):
This is awesome! Way less brittle than my original methodology and more easily updated going forward. I just had a few suggestions around documentation and a suggestion on the `util.aggregate_over_origin` function, and one blocker when using the other implementation of the `util.aggregate_over_origin` function, but otherwise great.

scripts/income_limits/30c_eligibility.py
```diff
@@ -0,0 +1,76 @@
...
+
+# write out table with all zctas x tracts to allow for tract lookup and
+# custom zip logic on front end
+fpath_out = util.DATA_FPATH / 'processed' / '30c_eligibility_by_tract_zcta.csv'
+print(f"Writing to {fpath_out}")
+eligibility_30c_by_tract_zcta.to_csv(fpath_out, index=False)
```
oyamauchi, line 72: This requires the runtime to do the aggregation over tract; is it intentional that there's no table that's one row per tract? (It's not a heavy lift for the runtime to do this; I'm just commenting that it's not what's specified in the Notion doc.)

oyamauchi:
Also, there are 5 rows in `ami_by_zcta` where the numbers are blank except for `ami_100` and `ami_150`, any idea what's up with those?

01003, 01701, 02325, 02760, 02763 (all in MA)

mikivee:
> * I'm not sure exactly what the data checks would be, but if it's not too much trouble, sounds like a good idea. Assert statements are fine.
 
I think actually the kind of manual checks you did on number of rows in the output is probably the appropriate level of checks here, since I can't know how many counties/countysubs/FMR areas there will be in the future. So I'm gonna not add any asserts. I think I've laid out the assumptions pretty clearly in the Notion/comments so should be easy to sort out in the future.

scripts/income_limits/30c_eligibility.py
```diff
@@ -0,0 +1,79 @@
...
+eligibility_30c_by_tract.to_csv(fpath_out, index=False)
+# write out table by zcta with min typeII error logic already applied
+fpath_out = util.DATA_FPATH / 'processed' / '30c_eligibility_by_zcta.csv'
+print(f"Writing to {fpath_out}")
+eligibility_30c_by_zcta.to_csv(fpath_out, index=False)
+# write out additional table by (zcta, tract)
+# to allow for lookup by zcta with custom agg logic applied at runtime
+fpath_out = util.DATA_FPATH / 'processed' / '30c_eligibility_by_tract_zcta.csv'
+print(f"Writing to {fpath_out}")
+eligibility_30c_by_tract_zcta.to_csv(fpath_out, index=False)
```
oyamauchi, line 79: Since we're not going to use this file at runtime, I don't think it's worth writing / checking in. If it turns out we need it, it'll be easy to update this script to add this code back in.

scripts/income_limits/util.py
```diff
@@ -0,0 +1,350 @@
...
+    "VI": "78"
+}
+# Note that while PR is a territory, census data for PR is stored
+# with the other states and DC so it does not require special handling
+TERRITORY_POSTAL_CODES = ['AS', 'GU', 'MP', 'VI']
+
+# HUD API
+with (pathlib.Path() / 'scripts' / 'income_limits' / 'config.json').open('r') as f:
+    config = json.load(f)
+HUD_API_HEADERS = {"Authorization": f"Bearer {config['hud_api_key']}"}
```
oyamauchi, line 76: Could you document this in the readme? Mention the file path, show the structure: `{"hud_api_key": "..."}`, and say RA employees should get the key from 1Password.

oyamauchi (approved):
This looks good to me; just two minor inlines. And @tsgra will need to approve as well.

mikivee:
> Also, there are 5 rows in `ami_by_zcta` where the numbers are blank except for `ami_100` and `ami_150`, any idea what's up with those?
> 
> 01003, 01701, 02325, 02760, 02763 (all in MA)

Oops sorry forgot to address this. I believe these correspond to the 11 geoids that failed in the get request for pulling AMI 80 and 150. See error log, where almost all are MA (state=25). I mentioned in L41-42 of the ami.py script that this would lead to only having AMI100 and AMI150 for a few county subs.  In the case that all of the countysubs overlapping the zcta failed to pull, we would get an aggregate of Null when mapping to zcta.

oyamauchi:
Discussed offline: for the zips with missing columns, we're going to approximate the missing columns with multiples of `ami_100`. (This is not how it works in general, but it's a good enough approximation in this edge case.)

Relatedly, there also appear to be a small number of tracts missing from `ami_by_tract`, which I think we can investigate separately, as a followup. (Currently no client of the API is using the `address` param, which is the only way tract-based lookup would be used.)

mikivee:
> Discussed offline: for the zips with missing columns, we're going to approximate the missing columns with multiples of `ami_100`. (This is not how it works in general, but it's a good enough approximation in this edge case.)
> 
> Relatedly, there also appear to be a small number of tracts missing from `ami_by_tract`, which I think we can investigate separately, as a followup. (Currently no client of the API is using the `address` param, which is the only way tract-based lookup would be used.)

Most recent commit fixes missing rows or entries in the AMI table. Also added a few data checks to flag if this happens in the future.

scripts/income_limits/ami.py
tsgra: Gotcha, my python version was out of date

tsgra (approved):
Looks good!

PR #385, opened 2024-03-22
Split authorities.json into a file for each state

oyamauchi:
## Description

We're about to populate authorities for every state's utilities in one
go, and that would make authorities.json unmanageably large. Since its
top-level key is already states, that makes it easy to split into one
file per state.

There's basically no downstream effect in the API server;
`authorities.ts` just recreates the same map that was previously in
the unified authorities.json. The scripts side is actually simplified
a bit.

The `state` key in a state's authorities is now optional, because we'll
be generating files with only a `utility` key.

https://app.asana.com/0/1206661332626418/1206882125337116

## Test Plan

`yarn lint` and `yarn test`.

Followed the new-state process (running generate-utility-data and then
authority-and-program-updater) for CA to make sure it goes smoothly.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -34,20 +34,9 @@ Filling out an entry for `incentive-spreadsheet-registry.ts` consists of creatin
 - Optionally declaring the header row number, if not the top row of the spreadsheet, in `headerRowNumber`
 - Optionally naming a filepath where _collected_ incentives will be written in `collectedFilepath`. This is experimental and affects how some of the scripts work, so you shouldn't do this for now unless you know what you're doing.
 
-First, **edit `authorities.json` manually** to include a top-level entry for
-the state you're adding, if it's not already present, like this:
+First, create a subdirectory in `data/` named with the new state's abbreviation.
 
-```json
-{
-  "<state-abbreviation>": {
-    "state": {},
-    "utility": {}
-  },
-  ...
-}
-```
-
-Then, run [`generate-utility-data.ts`](generate-utility-data.ts) to populate the list of utilities for the state in the authorities file. See [below](#utility-data) for details on that.
+Then, run [`generate-utility-data.ts`](generate-utility-data.ts) to populate the list of utilities in the state's authorities.json file. See [below](#utility-data) for details on that.
 
 [`generate-misc-state-data.ts`](generate-misc-state-data.ts) adds values to ancillary files to reflect the programs and authorities that will be needed for the JSON. This needs to happen first because our data schemas actually require an incentive's program/authority to be one of the listed members, and if that's not the case, the incentive will fail validation.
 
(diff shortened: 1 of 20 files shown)


oyamauchi:
The changes in `data` are all mechanical; no need to look at them closely. I just ran this:

```ts
import fs from 'fs';

const authoritiesByState = JSON.parse(
  fs.readFileSync('data/authorities.json', 'utf-8'),
);

for (const [state, auths] of Object.entries(authoritiesByState)) {
  fs.writeFileSync(
    `data/${state}/authorities.json`,
    JSON.stringify(auths, null, 2) + '\n',
  );
}
```

scripts/README.md
```diff
@@ -79,7 +68,7 @@ To encode relationships between incentives, see the [`relationships-README`](htt
 
 ## Utility Data
 
-`generate-utility-data.ts` reads a [dataset](https://downloads.energystar.gov/bi/portfolio-manager/Public_Utility_Map_en_US.xlsx) published by ENERGY STAR to create a mapping from ZIP codes to utilities. It writes to a CSV file in `scripts/data`, which is then imported into the SQLite database by `build.sh`. It also modifies `authorities.json` to include the utility IDs and names. This data is used in the `/api/v1/utilities` endpoint.
+`generate-utility-data.ts` reads a [dataset](https://downloads.energystar.gov/bi/portfolio-manager/Public_Utility_Map_en_US.xlsx) published by ENERGY STAR to create a mapping from ZIP codes to utilities. It writes to a CSV file in `scripts/data`, which is then imported into the SQLite database by `build.sh`. For any state with a subdirectory in `data`, it also modifies that state's `authorities.json` to include the utility IDs and names. This data is used in the `/api/v1/utilities` endpoint.
```
matelau, line 71: This isn't specific to the authorities changes you are making here, but more for my own edification and since you are updating the relevant README portions. When I was running this utility-data script locally I kept running into this error. 
image

Or this error if I ran the script from outside the build dir.
image


Rather than modifying the code I simply copied over the authorities.json and zip-to-utility.csv files to the build dir to make the script happy, but I was wondering if that is expected and if it should be part of the README?

oyamauchi: I think this is because you're compiling explicitly with `tsc`, rather than using `ts-node`, which is the expected workflow. The compiled files go into the `build` directory, which messes up script-relative paths.

datu925: I wasn't aware that `ts-node` is our expected way to run scripts. I've always compiled with `tsc` and I know some other folks run with `tsx`.

I'm happy to standardize, but if so, I think we should document this in the top-level readme and/or the `scripts` readme. Or if we don't want to standardize, we should figure out a way a more robust way to specify the paths.

scripts/lib/authority-and-program-updater.ts
```diff
@@ -139,17 +128,15 @@ export function updateAuthorities(
...
+      json[authority.authority_type.toLowerCase()] = {};
     }
-    json[stateUpper][authority.authority_type.toLowerCase()][authorityShort] = {
+    json[authority.authority_type.toLowerCase()][authorityShort] = {
       name: authority.name,
     };
   }
 
-  json[stateUpper].utility = existingUtilities;
+  json.utility = existingUtilities;
```
datu925, line 139: nit (and unrelated to this change): let's move this line further up in this method, since it's a bit odd to have the non-utility authorities logic sandwiched in between. I think either before the partitioning or after the error-checking are both fine.

datu925 (approved):
Minor comments; nothing blocking. LGTM.

PR #386, opened 2024-03-22
Add more_info_url to the v1 API; make program_url required

oyamauchi:
## Description

Putting links to our info about the IRA programs in `item.url` doesn't
make conceptual sense. This adds an optional `more_info_url` field at
the top level of incentives, and makes `program_url` required.

To make `program_url` required, I had to add program URLs to the IRA
programs in the data. There are official IRS pages for all the tax
credits, so I used those. The rebate programs don't have official
pages, so I just reused the URLs we're using as the more-info URLs on
those incentives.

https://app.asana.com/0/1206661332626418/1206843501113812

## Test Plan

`yarn test`

### DIFF
--- a/src/data/programs.ts
+++ b/src/data/programs.ts
@@ -48,8 +48,6 @@ const ira_programs = {
     },
     url: {
       en: 'https://www.irs.gov/credits-deductions/used-clean-vehicle-credit',
-      // TODO this page's content isn't translated into Spanish, but this is
-      // the URL where it should appear when they get it translated
       es: 'https://www.irs.gov/es/credits-deductions/used-clean-vehicle-credit',
     },
   },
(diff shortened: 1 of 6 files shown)


src/data/programs.ts
```diff
@@ -26,42 +26,76 @@ const ira_programs = {
...
+    },
   },
   creditForPreviouslyOwnedCleanVehicles: {
     name: {
       en: 'Federal Credit for Previously-Owned Clean Vehicles (25E)',
       es: 'Crédito para vehículos limpios de propiedad anterior (25E)',
     },
+    url: {
+      en: 'https://www.irs.gov/credits-deductions/used-clean-vehicle-credit',
+      // TODO this page's content isn't translated into Spanish, but this is
```
rpartridge, line 51: looks translated to me?

rpartridge (approved):

src/data/programs.ts
oyamauchi: Oh wow, they definitely just published that in the last few days.

PR #388, opened 2024-03-25
Add PDF-parsing and enable pdfs able to be hashed.

mmezher:
These are all library functions and aren't attached end to end yet, but I will leave a doc outlining how things should fit together and next steps.


scripts/check-data-freshness.ts
```diff
@@ -86,3 +123,18 @@ test('All URLs linking to current programs have an OK response code', async tap
     }
   }
 });
+
+test('PDF returned readable text, able to be parsed', async tap => {
+  // Check that PDF link can be read and returned.
+  // This PDF may oneday no longer exist, and if this test fails, the url should be checked first.
```
datu925, line 129: typo: one day

I feel a little weird about having a test that relies on a single PDF being available. Is this because you want there to be something exercising the code but we can't rely on the test working for all PDFs because of the timestamp issue?

scripts/check-data-freshness.ts
```diff
@@ -86,3 +123,18 @@ test('All URLs linking to current programs have an OK response code', async tap
     }
   }
 });
+
+test('PDF returned readable text, able to be parsed', async tap => {
```
datu925, line 127: optional: "PDF returned parsable text"

package.json
```diff
@@ -24,6 +25,7 @@
     "fastify-sqlite": "^1.1.0",
     "geocodio-library-node": "^1.6.1",
     "lodash": "^4.17.21",
+    "pdf-parse": "^1.1.1",
     "qs": "^6.11.2",
```
datu925, line 29: Actually, should these and cheerio be devDependencies instead of actual dependencies? AFAIK they're not needed for the deployed API.

I guess I don't know how GitHub Actions are configured and whether this distinction matters to those.

scripts/check-data-freshness.ts
```diff
@@ -32,6 +33,42 @@ async function checkUrlDataAvailability(
   return content.status;
 }
 
+// Returns the response data obtained by trying to request the URL content.
+async function returnAvailablePdfUrlData(
```
datu925, line 37: My gut tells me this at least some parts of this method should be consolidated with the non-PDF fetching above. Are there compelling reasons not to do that?

scripts/lib/website-comparisons.ts
```diff
@@ -33,17 +33,19 @@ export function isPdf(link: string, response: string): boolean {
   return false;
 }
 
-// Given a website link and its html, return a hash of the cleaned website. Do not return a hash if the site is for a PDF.
+// Given a website link and its data, return a hash of the cleaned website. If it's a PDF, attempt to convert to readable text first.
```
datu925, line 36: This comment is a bit confusing because if it's a PDF, it's *already* been converted to readable text.

scripts/lib/website-comparisons.ts
datu925, line 1: In general, I feel like this would be easier to understand if there were 3 steps that operate ~the same between PDF and HTML:

1. fetch the web content (maybe need a param to add the pdf-specific header)
2. parse the response
3. do something with the response (e.g. hash and save)

Right now the method structure doesn't reflect that – can we refactor to something closer/more analogous between PDF and HTML?

scripts/lib/website-comparisons.ts
```diff
@@ -33,17 +33,19 @@ export function isPdf(link: string, response: string): boolean {
...
 export function checkWebsiteAndReturnHash(
   link: string,
   response: string,
 ): string | null {
+  const hash = createHash('sha256');
   if (isPdf(link, response)) {
-    return null;
+    // This "response" is actually pre-parsed PDF text, via 'returnAvailablePdfUrlData', so can skip cleaning and just update hash.
```
datu925, line 43: See file comment above – this seems like a method structure smell.

datu925 (changes requested):

scripts/check-data-freshness.ts
mmezher: I actually found that the timestamp issue isn't a problem for this! It was surprising when I read over the text returned as actually just the text available in the PDF. The pdf-parse library was working successfully to scrape out the text in the file - what didn't work was the axios request for the PDF (with the wrong content header type and array buffer response not being properly defined). Post-that, I've found the PDF parsing library to work well at only returning the text (albeit, this doesn't help table-scraping the PDF files, since the text returned is pretty unformatted, but it's  a great improvement to before). 


For the test itself - I don't have a great idea of how to test this other than using a real web PDF, but from the overall suggestions you've provided below, I think that I can just ensure for PDF content, real content is returned, so this will become a non-issue.

oyamauchi: We should at least use a PDF hosted on one of our domains, if we're going to have a live-network-fetch automated test.

PR #390, opened 2024-03-25
Address PA QA findings and Duquesne feedback

rpartridge:
Fixing QA issues that came up during the testing party today, as well as some feedback we got from our contacts at Duquesne who have been reviewing the incentives spreadsheet. 

Main changes:
- Add 2 additional EV incentives. Duquesne emailed and wanted these included. Technically 1 incentive but we need to list twice, once for new and once for used EV
- Originally I had all PA incentives listed as applies to both homeowner and renter because none of the programs specified homeowner only, so I assumed both were eligible. However Hannah brought up [this mapping](https://github.com/rewiringamerica/incentive-data-llm/blob/main/src/constants.ts#L16) that we can use, so in this PR I've updated to be consistent with that mapping
- Added ® to all references to Energy Star
- Updated a couple weatherization program names to be more specific. This was a little confusing because there's a state-wide free weatherization assistance program and a few utility ones, so I just added the authority name
- Couple other changes reported by our Duquesne point of contact in the spreadsheet, one for incentive amount for PA-9 and one for incentive link for PA-2

### DIFF
--- a/data/PA/incentives.json
+++ b/data/PA/incentives.json
@@ -13,8 +13,7 @@
       "number": 1
     },
     "owner_status": [
-      "homeowner",
-      "renter"
+      "homeowner"
     ],
     "short_description": {
       "en": "Free energy audit and weatherization services for income-qualified residents. Average value per household is $7,669 depending on results.",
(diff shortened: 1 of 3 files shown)


data/PA/incentives.json
```diff
@@ -206,12 +198,11 @@
...
     },
     "owner_status": [
-      "homeowner",
-      "renter"
+      "homeowner"
     ],
     "short_description": {
-      "en": "$150 rebate per unit for central air conditioner.",
-      "es": "Reembolso de $150 dólares por unidad de aire acondicionado central"
+      "en": "Up to $250 rebate per unit for central air conditioner.",
```
oyamauchi, line 204: This doesn't match `amount.number`

oyamauchi (approved):
Yeah, we'll want to do a global find/replace on ENERGY STAR®, but I'd like to wait until we have JSON-to-spreadsheet working so that JSON is the source of truth and we can do the change there.

PR #393, opened 2024-03-26
MI data refresh updates

rpartridge:
Data refresh for Michigan. Main changes:
- Combine a few rebates, for example heat pumps of different efficiencies. It seems like that's the direction we're moving towards
- Update with 2024 programs


data/MI/incentives.json
```diff
@@ -183,19 +183,17 @@
       "rebate"
     ],
     "item": "smart_thermostat",
-    "program": "mi_consumersEnergy_residentialHeating,CoolingAndWaterHeatingProgram",
+    "program": "mi_consumersEnergy_residentialApplianceRebateProgram",
```
rpartridge, line 186: new 2024 program, but I updated the existing 2023 incentive

data/MI/incentives.json
```diff
@@ -465,73 +462,15 @@
...
-    "authority": "mi-dte",
-    "payment_methods": [
-      "rebate"
-    ],
-    "item": "heat_pump_air_conditioner_heater",
-    "program": "mi_dTE_airConditioners",
-    "amount": {
-      "type": "dollar_amount",
-      "number": 750
+      "number": 850,
```
rpartridge, line 465: combined MI-22, 23, 24 into this incentive and made it a range. they're all heat pump rebates for various efficiencies

data/MI/incentives.json
```diff
@@ -1061,7 +978,7 @@
       "homeowner"
     ],
     "short_description": {
-      "en": "$1500 rebate for ASHP – Central Air-Source Heat Pump (Must be listed on ashp.neep.org)."
+      "en": "Up to $1500 rebate for central air source heat pump. Rebate amount depends on appliance efficiency. Requires a licensed mechanical contractor."
```
rpartridge, line 981: combined MI-50 into MI-49, both types of air source heat pumps

data/MI/incentives.json
```diff
@@ -1084,34 +1001,12 @@
       "homeowner"
     ],
     "short_description": {
-      "en": "$900 rebate per outdoor unit for mini/multi split air-source heat pumps. Limit 2 units."
+      "en": "$900 rebate per outdoor unit for mini/multi split air-source heat pumps. Limit 2 units. Additional $100 per indoor unit."
```
rpartridge, line 1004: combined MI-52 into MI-51. 51 was the additional $100 per indoor unit but I think they can be combined?

data/MI/incentives.json
```diff
@@ -1278,49 +1173,49 @@
...
     ],
     "short_description": {
-      "en": "Get rebates up to $155 on smart thermostats via the I&M Marketplace."
+      "en": "$25 rebate for smart thermostats and $15 rebate for Wi-Fi thermostats. Limit 2 of each kind per account lifetime. Household must have electric heat."
     },
     "bonus_available": true
   },
   {
-    "id": "MI-61",
+    "id": "MI-65",
```
rpartridge, line 1189: this diff is a little confusing. changes:
1. omit MI-61-63. I can't find any reference to them on I&M's website so I think they must have expired in 2023
2. added a couple new incentives

oyamauchi (approved):
I haven't quite decided what I want to do re: dates in the API. Before anything else we definitely need to audit all the ones that are marked as ending in 2023, and likely have some kind of ongoing monitoring for incentives with end dates in the past (along the lines of the data freshness checks). Then the question is: should the API filter out past-ending incentives by default? That part I don't know.

I'm not sure about deleting expired incentives entirely. I'm pretty sure it's a non-goal for us to keep accurate data about past incentives, so in principle I think we should delete. But then there are just procedural questions like:

- Should we keep expired incentives around in the data for a little while (how long?) without serving them in the API, in case the programs are revived?
  - I feel like the answer should be no, as long as we don't have knowledge that a program _will_ be revived; if a program is revived, it's not a huge lift to enter it as a new row
- Should IDs be reusable?
  - My inclination is no just on principle, even though AFAIK there currently would be no negative consequences to reusing. This may be tricky to enforce in spreadsheets, though. It's not as simple as "use the lowest number that's greater than all currently in the sheet": what if we add "NY-78", then delete it before adding any others, then someone goes to add a new row?

I'll raise these questions in the incentives team meeting, I think.

data/MI/incentives.json
```diff
@@ -1329,29 +1224,116 @@
       "homeowner"
     ],
     "short_description": {
-      "en": "$100 rebate on an Enel X Level 2 EV Charger."
-    }
+      "en": "$100 rebate for electric resistance (non-heat pump) water heater. Must be high durability, point-of-use electric, or whole home electric."
```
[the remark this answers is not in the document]
rpartridge: so originally I was marking non-heat pump water heaters and non-heat pump clothes dryers as `other`, but during one of the QA meetings it came up that that was filtering these incentives out of the embed, [which we don't want](https://rewiringameri-g3x1100.slack.com/archives/C04E4THEGMT/p1710224433812529?thread_ts=1710186110.616129&cid=C04E4THEGMT). it's a little hacky to mark them as heat pump just to get them to show up in the calculator but IMO seems preferable to excluding them - is there a better option?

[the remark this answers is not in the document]
oyamauchi: OK it just dawned on me that we do have the item values `non_heat_pump_clothes_dryer` and `non_heat_pump_water_heater` in the API. (I don't know if they're in the spreadsheet templates though?)

Anyway we should definitely use those values for those items, and if we want them to be shown in the frontend, it's a very simple frontend change.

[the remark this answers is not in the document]
rpartridge: oh will use that then!

PR #394, opened 2024-03-26
Account for Colorado income tax in state tax credit eligibility/maximums

veekas:
## Description

Colorado offers tax credits against state taxes. This PR does as we do with federal taxes and calculates the user's tax burden to avoid showing tax credit incentives if the user wouldn't be able to claim them because they don't owe enough tax.

The state has a flat tax rate of 4.4% and does not have a standard deduction.

[Asana](https://app.asana.com/0/1205271193902672/1206661332626430/f)

## Test plan

- added tests! `yarn test`


src/lib/state-incentives-calculation.ts
```diff
@@ -129,6 +136,21 @@ export function calculateStateIncentivesAndSavings(
...
+      typeof stateTaxOwed.tax_owed === 'number' &&
+      stateTaxOwed.tax_owed < item.amount.number
+    ) {
+      // mark the incentive ineligible if the user owes no taxes and the incentive isn't redeemable in another way
+      if (stateTaxOwed.tax_owed === 0 && item.payment_methods.length === 1) {
+        eligible = false;
+      }
+
+      // change the tax credit amount to match the user's max tax liability
+      item.amount.number = stateTaxOwed.tax_owed;
```
oyamauchi, line 151: I'm a little hesitant about this. AFAIK we're currently always presenting incentives with non-customized amounts (though correct me if I'm wrong!) so this would be breaking new ground.

It could result in showing incentives with weird non-round numbers, and it also might not be clear to the user why we're showing an amount that's different from the amount in the incentive's official definition.

It's also sort of an incomplete version of what it's trying to do: suppose you're eligible for two tax credits of $1000 each, and your tax liability is $800. This will set both tax credits' amounts to $800, and you'll see that you're eligible for two $800 credits. But you're still only going to be able to claim $800 worth of credits total, not $1600.

I do think it's valuable to _somehow_ communicate when a user can't actually claim the total value of all the tax credits we show them, but I think it needs to be a deeper rethink.

oyamauchi (changes requested):
Looks mostly good, just a couple of things.

src/lib/state-incentives-calculation.ts
veekas: Yeah, that's fair. I'll revert it!

src/lib/state-incentives-calculation.ts
```diff
@@ -199,9 +200,23 @@ export function calculateStateIncentivesAndSavings(
...
+  // Get state tax owed to determine max potential tax savings
+  const stateTaxOwed = estimateStateTaxAmount(
+    request.household_income,
+    stateId,
+  );
+
+  // You can't save more than tax owed. Choose the lesser of state tax owed vs tax credit savings
+  const maxTaxCreditSavings = stateTaxOwed
+    ? Math.min(stateTaxOwed.taxOwed, savings.tax_credit)
+    : savings.tax_credit;
```
oyamauchi, line 212: I think you can just update `savings.tax_credit` in place and avoid the part down below

oyamauchi (approved):

PR #396, opened 2024-03-27
Refactor website check to return full response

mmezher:
This will enable us to use the same function to read both HTML and PDF website data in subsequent PR.


mmezher:
Actually, going to move review to the other PR, whose only other change is moving the function into website-comparisons.ts

PR #397, opened 2024-03-27
Move checkUrlDataAvailability to website-comparisons.ts

mmezher:
This better separates the status check test from the functions that will eventually be used for multiple checks for PDF and HTML content.

### DIFF
--- a/scripts/check-data-freshness.ts
+++ b/scripts/check-data-freshness.ts
@@ -1,36 +1,6 @@
-import axios from 'axios';
-import axiosRetry from 'axios-retry';
 import { test } from 'tap';
 import { PROGRAMS } from '../src/data/programs';
-
-// Returns the status obtained by trying to request the URL content.
-async function checkUrlDataAvailability(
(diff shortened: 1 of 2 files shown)


scripts/check-data-freshness.ts
```diff
@@ -53,7 +23,19 @@ test('All URLs linking to current programs have an OK response code', async tap
         console.error('${url_to_check} is not a valid URL for program: ', key);
         process.exit(1);
       } else {
-        const status = await checkUrlDataAvailability(url_to_check);
+        const response = await returnAvailableUrlData(
```
datu925, line 26: Optional rename (and you can do this as a follow-up when this is all over if it's easier) to something like `retrieveWebContent`.

datu925 (approved):

scripts/check-data-freshness.ts
mmezher: Very cool suggestion, thanks! Done

PR #398, opened 2024-03-27
Add formatting and header for Google sheet export.

datu925:
This gets the output closer, though still not identical, to our template. For comparison:

[Colorado data](https://docs.google.com/spreadsheets/d/1nITjSNRWJjSusB0fWuSS63fqynm_4Co7KEeuo9Cm7fU/edit#gid=30198531)
[Sample export of CO data](https://docs.google.com/spreadsheets/d/1CmJWdKTG0HF49PcdV5fRUi9N-IFJeY-uAvonKPoCUG0/edit#gid=0)

Known limitations:
- The program description header doesn't have a link. I'm inclined to either hard-code that or just not support links in the headers. We can definitely have them in the data model sheet.
- The header groupings (Rebate Amount Fields and Restrictions / Eligibility Conditions) are hard-coded by column number. I also thought about having the user specify a start and end column by name. I think we can revisit it if we notice this breaking a lot.
- links/URLs still don't become clickable by default. I will address this in a separate PR

To make this easier, I separated the previous collectedIncentiveToGoogleSheet method into two methods. One performs the value transformations to get to something that any spreadsheet needs; the other is concerned with exporting and formatting specifically to Google Sheets.

The testing story is a bit inconvenient here. We can absolutely add another test that verifies that the standard header output is expected, but it will be a very large golden file. Even the foo-bar-baz version without the header is pretty large.


scripts/lib/google-sheets-exporter.ts
```diff
@@ -0,0 +1,219 @@
...
+            {
+              startIndex: 22,
+              format: {
+                link: {
+                  uri: 'https://docs.google.com/document/d/1nM9320uOUYpfpD3eZ4DFo32Scd5kauLEL17j1NYB5Ww/edit#bookmark=id.s9a0j7v2m6vb',
+                },
+              },
+            },
+            {
+              startIndex: 32,
```
oyamauchi, line 147: These two `startIndex` values should use the constants defined up top

oyamauchi (approved):
Nice. I think there's no need to go to great lengths to have links in headers. Whatever the links are can go in the separate explanatory sheet.

I assume data validations for the enum columns are also on your radar as a followup?

scripts/lib/google-sheets-exporter.ts
datu925: These are actually unrelated – they are the start/end indexes of `guidelines` in `Rebate Amount Fields (guidelines)` to properly display the hyperlink (that one's easy enough since those fields are hard-coded anyway).

datu925: I'll add a comment to make that clearer.

datu925: Actually I decided to do some string parsing – less brittle than the hard-coded indices (though only slightly).

datu925:
> Nice. I think there's no need to go to great lengths to have links in headers. Whatever the links are can go in the separate explanatory sheet.
> 
> I assume data validations for the enum columns are also on your radar as a followup?

Yes, planning to get to that tomorrow!

PR #400, opened 2024-03-27
Add link formatting to Google Sheet exports.

datu925:
Links in Google Sheets don't autoformat (e.g. become clickable) when written via API unless the formatting data is explicitly supplied. This is harder for us to do up-front (though is a valid alternative if we don't like this approach).

Instead, once we've written the sheet, we can simply use the API to read cell values and write them back as if the user had entered them. This triggers Google Sheets to format the input. Unfortunately I don't believe this is possible to do as part of initial spreadsheet creation.

For an alternative, we'd need a reliable URL detector regex and we'd have to scan through all the cells, finding links and then adding format data to each cell indicating where the link should be placed. That seemed harder overall since URL detection is probably a rabbit hole.

I see a few ways we can adapt this once we're ready to handle the description word/character count columns, so I'm not too worried about that.

Tested manually – since this is all about invoking APIs, there's not much to test on our end. See https://docs.google.com/spreadsheets/d/1MSUJRFNAmO_iqnU_R-tMv_v9FWGVtQzUDW-29ZSxmNk/edit#gid=479908268 for an example of clickable links.

### DIFF
--- a/scripts/export-to-google-sheets.ts
+++ b/scripts/export-to-google-sheets.ts
@@ -7,6 +7,8 @@ import { authorize } from './lib/auth-helper';
 import { collectedIncentiveToGoogleSheet } from './lib/format-converter';
 import { FIELD_MAPPINGS } from './lib/spreadsheet-mappings';
 
+const INCENTIVE_SHEET_NAME = 'Incentives Data';
+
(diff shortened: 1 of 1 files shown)


oyamauchi (approved):
That's a silly hack, but great that it works!

PR #405, opened 2024-03-28
Refactor value mappings into field metadata.

datu925:
This is a refactor prior to using it for more export-to-google work, but it's worth doing in its own right.

We don't have any type-safety for the values, but that's no different than before.

As part of doing this, I added values for filing_status since those were missing, plus the value descriptions.

Testing: prior to deleting the old version of VALUE_MAPPINGS, I logged both versions to the console and they had zero diffs.

### DIFF
--- a/scripts/lib/spreadsheet-mappings.ts
+++ b/scripts/lib/spreadsheet-mappings.ts
@@ -23,6 +23,11 @@ export type FieldMetadata = {
   // the column name that is used.
   column_aliases: string[];
   description: string;
+  // To provide aliases for enum values and generate explanations.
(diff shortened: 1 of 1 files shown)


scripts/lib/spreadsheet-mappings.ts
```diff
@@ -23,6 +23,11 @@ export type FieldMetadata = {
   // the column name that is used.
   column_aliases: string[];
   description: string;
+  // To constrain enum values and generate explanations.
+  // We might be able to constraint the values below with some more type magic.
```
mmezher, line 27: nit: constrain the values below (remove t)

mmezher (approved):

PR #406, opened 2024-03-28
Add Standardized Enum List Values sheet to Google export.

datu925:
This has a previous PR as a base, but that's just a refactor to get our metadata in a more logical configuration.

[Sample output](https://docs.google.com/spreadsheets/d/1jXxinYfxFqlnfuTa8uTfObi3rX28-tC2hV01KgHmyJA/edit?usp=sharing)

Notable diffs:
1. The new version has a definition column even if there are no values with definitions. This frankly didn't seem worth adding the conditionals/complexity for, but it is certainly possible if needed.
2. Some of the values are different because the current enums don't match our schema. For example, amount type only has 3 values, whereas our spreadsheet enums allowed many more. I never understood why we had some many extra variations in the spreadsheet, since they are derivable from the other columns anyway (e.g. percent with a cap just means the amount type is percent and a maximum is populated).

Not in this PR:
- protecting the sheet from editing
- using data validation from the main Incentives Data tab


scripts/lib/google-sheets/standardized-enums.ts
```diff
@@ -0,0 +1,142 @@
...
+}): sheets_v4.Schema$Sheet {
+  const rowData: sheets_v4.Schema$RowData[] = [];
+
+  // Filter to the metadata where values is defined.
+  const enumListsToWrite: OrderedValuesMetadata[] = Object.values(metadata)
+    .filter(field => field.values)
+    .map(field => {
+      return {
+        columnName: field.column_aliases[0],
+        values: Array.from(Object.values(field.values!)),
```
oyamauchi, line 72: `Object.values` is an array already

oyamauchi (approved):

PR #408, opened 2024-03-29
Add formula-based description word/char count columns.

datu925:
Two related but separate changes:

1. introduce an argument to drop columns when reading them in in the flatToNested method.
2. introduce an argument to add the formula columns when writing them. They're hard-coded to the word/char count formulas, but the approach should work for other formulas if we need to add them in the future. It works because Google Sheets API has a valueRenderOption param that recognizes formulas when reading spreadsheet data.

The API is a bit awkward because it's not clear from the name `descriptionColumnName` that it would control formulas. If you prefer, we can introduce an additional boolean param called `addDescriptionFormulaColumns`, and if it's true, we require the column name as well.

Testing: unit tests, and ran on Colorado to make sure that we were dropping the columns on the way in, and adding them back on the way out. Output: https://docs.google.com/spreadsheets/d/1A18zvzRtVYiOghdRqCGBd_DWUGmrx7bgdsdItocklIw/edit#gid=0

### DIFF
--- a/scripts/lib/google-sheets/incentives-sheet-exporter.ts
+++ b/scripts/lib/google-sheets/incentives-sheet-exporter.ts
@@ -3,6 +3,7 @@ import util from 'util';
 import { CollectedIncentive } from '../../../src/data/state_incentives';
 import {
   SpreadsheetData,
+  colToLetter,
   collectedIncentivesToSpreadsheet,
 } from '../format-converter';
 import { AliasMap } from '../spreadsheet-mappings';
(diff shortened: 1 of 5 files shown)


scripts/lib/google-sheets/incentives-sheet-exporter.ts
```diff
@@ -95,22 +96,42 @@ export function collectedIncentiveToGoogleSheet(
   incentives: CollectedIncentive[],
   fieldMappings: AliasMap,
   useStandardHeader: boolean,
+  descriptionColumnName: string | undefined = undefined,
```
oyamauchi, line 99: The TS idiom for this is `descriptionColumnName?: string` -- it adds the `undefined` part implicitly

oyamauchi (approved):

scripts/lib/google-sheets/incentives-sheet-exporter.ts
datu925: Oops, not sure how I forgot about that.

PR #409, opened 2024-03-29
Add data validations and sheet protection.

datu925:
[Sample output](https://docs.google.com/spreadsheets/d/1lS5htTOE2Nji9Nqy_6HzA50dZuPEYukb_G1tUXe63ys/edit#gid=0)

Notable diff from current approach: I currently have it so that users can edit the protected sheet; it just shows a warning. In my experience, the sheet protection has been more trouble than it's worth - see Slack for lots of examples of people needing sheets unlocked. I think we want the validations to discourage changes but ultimately allow people to make changes if they think the schema needs to be updated.

It's an easy change to switch to actual protection if you prefer, in which case I would make engineering-team@rewiringamerica.org editors.

Not in this PR
- there are other protected ranges in our templates like the example data row or the IDs column, but I think they are also unnecessary, so I'm not inclined to implement them. It's pretty easy to add them, especially since they're hardcoded (row 3 and column 1)

### DIFF
--- a/scripts/lib/google-sheets/incentives-sheet-exporter.ts
+++ b/scripts/lib/google-sheets/incentives-sheet-exporter.ts
@@ -6,7 +6,7 @@ import {
   colToLetter,
   collectedIncentivesToSpreadsheet,
 } from '../format-converter';
-import { AliasMap } from '../spreadsheet-mappings';
+import { AliasMap, FieldMetadata } from '../spreadsheet-mappings';
 
 const REBATE_AMOUNT_FIELDS_START = 15; // inclusive
 const REBATE_AMOUNT_FIELDS_END = 22; // exclusive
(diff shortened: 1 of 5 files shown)


scripts/lib/google-sheets/incentives-sheet-exporter.ts
```diff
@@ -91,24 +92,62 @@ function createCellValue(
...
+): sheets_v4.Schema$DataValidationRule | undefined {
+  for (const [ind, field] of fieldsWithEnums.entries()) {
+    if (col === field.column_aliases[0]) {
+      const columnLetter = colToLetter(ind * 3 + 1);
+      const lastEnumRow = Object.keys(field.values!).length + 1;
+      return {
+        strict: true,
+        // Make Google Sheets to show a dropdown UI element. Unfortunately,
+        // it's not possible in the API to force the "chip" version, just a
+        // regular list.
```
oyamauchi, line 107: I can't see this in the sample sheet because I only have view access, but I believe you that it works

scripts/lib/google-sheets/incentives-sheet-exporter.ts
```diff
@@ -91,24 +92,62 @@ function createCellValue(
...
+        showCustomUi: true,
+        condition: {
+          type: 'ONE_OF_RANGE',
+          values: [
+            {
+              userEnteredValue: `='Standardized Enum List Values'!$${columnLetter}$2:$${columnLetter}$${lastEnumRow}`,
+            },
+          ],
```
oyamauchi, line 115: What if it's a potentially multi-valued column like payment methods?

oyamauchi (approved):
Approving, but wondering what the story is for columns that are "one or more from this set of values".

scripts/lib/google-sheets/incentives-sheet-exporter.ts
datu925: I changed permissions to editor in case you want to see it later.

scripts/lib/google-sheets/incentives-sheet-exporter.ts
datu925: It's an awkward case in the current state, and this PR doesn't change anything about it. We currently have a `Multiple` type in the enums, so I propagated that into the spreadsheet-mappings.ts. But I don't like that because it's not a valid value in the schema.

The spreadsheet-to-json flow is capable of processing and validating comma-delimited arrays, so this is mostly about finding a way to let the spreadsheet allow that. For the least protection, we have the current approach, where collectors can go modify the allowed values. That could be our official guidance – if you need to add a multi-valued value, just go augment the spreadsheet.

Or instead of a "one of range" data validation, we could use a custom formula validation. Something like: `=iserror(product(map(split(Y9,","), lambda(txt, match(trim(txt),Y2:Y7,0)))))`, where Y9 is the cell in question, and Y2:Y7 is the allowable values. Basically splitting by comma and running each value through the allowable list, then verifying all of them are found. I'm going to add that to my handoff doc, but I'll try to get to it if I have time.

PR #410, opened 2024-03-29
Remove item.url from v1

oyamauchi:
## Description

This is no longer used by the frontend, and the URLs now get
redirected to incentive-centered pages; I think the whole concept of a
URL for an item is dead.

v0 still uses the URLs, in the `more_info_url` field, so we can't
delete them. We can't even change them to the incentive-centric
`homes.rewiring` URLs, because these URLs are relative, and the host
is assumed to be `www`. So we'll be stuck with this until v0 can be deleted.

## Test Plan

`yarn test`. Make sure the embed frontend does not use this field.

### DIFF
--- a/src/routes/v1.ts
+++ b/src/routes/v1.ts
@@ -31,9 +31,6 @@ function transformIncentives(
     item: {
       type: incentive.item,
       name: t('items', incentive.item, language),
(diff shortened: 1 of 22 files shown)


oyamauchi:
Merging without review

PR #417, opened 2024-04-03
Add utility data for all states

oyamauchi:
## Description

Since we want to show a utility selector for everybody, we need
utilities for all the states. So here they are!

I did not QA these as intensively as I did the initial set of
states. I figure whoever adds each subsequent state's incentives
should do that, as I documented in the README.

I didn't Google every single utility. I looked for weird acronyms or
artifacts, or utilities that obviously seemed to be the same one under
different names. I made some changes to the name-cleaning logic to fix
a lot of the common cases automatically (which is why there are
changes to the GA and PA files).

The California file, and new overrides/exclusions for CA, are taken
from #380.

### Followups

- Remove the `include_beta_states` conditioning from `/v1/utilities`
  to allow utility selection in all states. (That maybe should wait
  until after the frontend has the "Other" option added.)

- Add territories. The Energy Star data doesn't cover them. However,
  fortunately, the utility situation in the territories is not super
  complicated.

https://app.asana.com/0/1206661332626418/1206843501113809

## Test Plan

Run the script to produce this data. Make sure running it again
doesn't change anything. QA the resulting data as described above.


scripts/lib/utility-data-overrides.ts
```diff
@@ -95,6 +192,60 @@ export const OVERRIDES = new Map<string | number, string>([
   [21048, 'Wyandotte Municipal Services'],
   [19125, 'Traverse City Light & Power'],
 
+  // MN
+  [150, 'Adrian Public Utilities'],
+  [1101, 'Bagley Public Utilities'],
+  [4577, 'Crow Wing Power'],
```
rpartridge, line 198: I like this creative utility name

scripts/generate-utility-data.ts
```diff
@@ -40,12 +40,27 @@ function convertName(
   cleaned = cleaned.replace(/\b,?\s+(I(nc)?|Co)\.?$/i, '').trim();
   // Remove "(for __ reporting)" parenthetical
   cleaned = cleaned.replace(/\(.*reporting.*\)/i, '').trim();
-  // Spell out "Pwr", "Assn", "Elec", and "Coop"
-  // The dataset is inconsistent about this
+  // Spell out "Pwr", "Assn", "Elec", "Coop", and "Dist". Replace some
+  // spaced-out abbreviations with un-spaced versions.
+  // The dataset is inconsistent about this.
```
rpartridge, line 45: nice 👍

rpartridge (approved):
I didn't look through all the state authority files, but went through the changes in generate-utility-data.ts and everything looks reasonable - I like the approach of fixing on a state by state basis. lgtm!

PR #419, opened 2024-04-04
Add Pennsylvania DEP EV rebate incentives

rpartridge:
Our Duquesne point of contact let us know about this program and asked if we can add these state level EV rebates. These rebates have a couple of permutations:
- They apply to new and used EVs (but slightly different criteria - for example, a new EV can be leased and still be eligible, used has to be purchased)
- They apply to battery EVs and plug in hybrids
- All of the incentives are income-qualified (400% of poverty level), but there are additional incentives available for residents at 200% of the poverty level.

I ended up adding 4 variations of this incentive to cover all of the cases:
PA-47 - New EV or PHEV (income qualified, 400% of poverty level)
PA-48 - Used EV or PHEV (income qualified, 400% of poverty level)
PA-49 - New EV or PHEV with bonus incentives (income qualified, 200% of poverty level). Users not eligible for PA-47 if they're eligible for this one
PA-50 - Used EV or PHEV with bonus incentives (income qualified, 200% of poverty level). Users not eligible for PA-48 if they're eligible for this one


data/PA/authorities.json
```diff
@@ -1,14 +1,17 @@
...
-    }
-  },
   "other": {
     "pa-first-energy": {
       "name": "First Energy"
     }
   },
+  "state": {
+    "pa-department-of-community-and-economic-development": {
+      "name": "Department of Community and Economic Development"
```
oyamauchi, line 9: Is there a particular reason this doesn't have the state name in it too?

oyamauchi (approved):
Ah this makes me realize we may to split `item` values into pure 
battery EVs versus PHEVs. Updated [my doc](https://www.notion.so/rewiringamerica/Granular-items-or-sub-items-274423aff1cd42f880367f13510500ae)

data/PA/authorities.json
rpartridge: no reason - will update

PR #420, opened 2024-04-04
Release PA incentives

rpartridge:
Launch 🎉 Moving forward this because:
- We're doing an event in PA on 4/22 and this needs to be ready by then
- PA utilities have started working on integrating the embed into their sites, and to avoid using the `beta_states` flag these need to be live


oyamauchi (approved):
Can you also look through `update-fixtures.sh` and remove the beta states argument and TODO comments for PA?

PR #421, opened 2024-04-05
[WAIT TO MERGE] Launch Colorado

oyamauchi:
## Descriptions

A surprising number of changes here:

- There were a bunch of incentives that still had "unknown" payment
  type. I looked at the utility websites and they're clearly
  advertised as rebates (post-purchase), so I updated the spreadsheet
  and synced that into JSON.

- I updated the program names to be more user-friendly when displayed
  in the incentive cards in the frontend. (This implicates a deeper
  question around how much we should strive to keep frontend concerns
  out of the API design -- this style of program name is pretty
  obviously driven by frontend concerns -- but given the imminent CO
  launch we're punting it for now and following what we've been doing
  in other states.)

  One program URL is updated too, because the old one was broken.

- The v0 endpoint has actually had the potential to pull in
  non-federal incentives from `calculateIncentives` all along, and
  only now did that come into play. (It didn't before because no
  `LAUNCHED` state had `pos_rebate`, `performance_rebate`, or
  `tax_credit` incentives and v0 could never have
  `include_beta_states` be true; it started now because CO has tax
  credits. Explicitly have it calculate federal incentives only.

- `incentives-relationships.test.ts` was importing `schemas.test.ts`,
  which actually caused the tests in `schemas.test.ts` to be run
  twice. Factor out the shared code into a new file, to stop that.

- We've been using 80212, a CO zip, all along as a generic "no state
  coverage" zip code for end-to-end tests. Those days are
  over. Switched it to 84106, a zip in Salt Lake City, UT with a
  similar AMI.

## Test Plan

`yarn test`. View the incentives with the calculator frontend pointed
at local server, to make sure the program names look good.

### DIFF
--- a/data/CO/incentives.json
+++ b/data/CO/incentives.json
@@ -5236,7 +5236,7 @@
     "authority_type": "utility",
     "authority": "co-san-isabel-electric",
     "payment_methods": [
-      "unknown"
+      "rebate"
     ],
     "item": "non_heat_pump_clothes_dryer",
     "program": "co_sanIsabelElectric_sanIsabelElectric",
@@ -5259,7 +5259,7 @@
     "authority_type": "utility",
     "authority": "co-san-isabel-electric",
     "payment_methods": [
-      "unknown"
+      "rebate"
     ],
     "item": "heat_pump_clothes_dryer",
     "program": "co_sanIsabelElectric_sanIsabelElectric",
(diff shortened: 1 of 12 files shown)


RandomEtc (approved):
This looks good to me, and I think could merge any time.

That said, might be safer to merge without promoting CO to launched, and then test one more time before landing a separate PR to switch to launched?

PR #425, opened 2024-04-09
MI QA fixes

rpartridge:
## Description 

Fixes for things that came up in the QA party on Friday https://docs.google.com/document/d/1jbSV3q1yEnDnrXJd5wKbdIRJk2yaGlBcjSkhBPW4aZQ/edit

The biggest change is that I noticed the expired incentives are still showing up - since the front end logic isn't there to exclude them yet, I marked all the expired incentives (there are a lot in MI 😢 ) as omit from API. These expired incentives were replaced with the new 2024 ones

## Open question

~@hblumberg flagged that we don't want to use mathematical symbols in descriptions as per the [program description guidelines](https://docs.google.com/document/d/1nM9320uOUYpfpD3eZ4DFo32Scd5kauLEL17j1NYB5Ww/edit#bookmark=id.mryjojvw3dsg). I agree with this for the most part but for this example (MI-71) the symbol is greater than or equal to which is kind of cumbersome to write out... do we think leaving in the symbol is ok in this case? (description: "Up to $600 rebate per heater for ENERGY STAR® heat pump water heaters. $600 for UEF ≥ 3.5, $475 for UEF ≥ 3.0.")~


hblumberg (approved):
Code changes LGTM! Will defer to @oyamauchi on the >= question. I've been saying things like "UEF must be at least .88", but I know we don't always have the characters!

oyamauchi:
Yeah there are a few ways to word it:

- `UEF of 3.5 or more`
- `UEF of at least 3.5`

Generally:
- `UEF > y` -> "UEF over y"
- `UEF < y` -> "UEF under y"
- `UEF >= y` -> "UEF of y or more"
- `UEF <= y` -> "UEF up to y"

PR #426, opened 2024-04-09
GA QA fixes and Spanish translations

rpartridge:
Addresses GA QA and adds Spanish translations.

QA fixes:
- splits out GA power heat pump incentives into 2 separate incentives, one for geothermal heat pumps and one for mini splits


oyamauchi (approved):
I do wish there were more than 9 incentives in Georgia, but it does make our lives a little easier

PR #428, opened 2024-04-10
Add UGI rebates and PECO EV rebate

rpartridge:
Couple of new rebates that PA DEP emailed about. See ticket for more details: https://app.asana.com/0/1206629812394927/1207049982896669/f

Additions:
1. UGI utilities heat pump and appliance rebates
2. PECO EV rebate

### DIFF
--- a/data/PA/incentives.json
+++ b/data/PA/incentives.json
@@ -1089,5 +1089,67 @@
     "start_date": "2023-07-01",
     "end_date": "2024-06-30",
     "low_income": "pa-pennsylvania-department-of-environmental-protection-bonus"
+  },
+  {
+    "id": "PA-51",
+    "authority_type": "utility",
+    "authority": "pa-ugi-utilities",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "heat_pump_air_conditioner_heater",
+    "program": "pa_uGIUtilities_equipmentRebateProgram",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 1250,
+      "minimum": 150,
+      "maximum": 1250
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "Up to $1,250 rebate for ENERGY STAR® certified mini-split heat pumps. Rebate amount varies depending on heat pump efficiency and capacity."
+    }
+  },
+  {
+    "id": "PA-52",
+    "authority_type": "utility",
+    "authority": "pa-ugi-utilities",
+    "payment_methods": [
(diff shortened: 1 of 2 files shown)


data/PA/incentives.json
```diff
@@ -1089,5 +1089,67 @@
...
+    }
+  },
+  {
+    "id": "PA-53",
+    "authority_type": "utility",
+    "authority": "pa-peco",
+    "payment_methods": [
+      "rebate"
+    ],
+    "item": "new_electric_vehicle",
```
oyamauchi, line 1142: Does this apply to used EVs too?

If so, I'm actually not sure we should duplicate the incentive with `item: used_electric_vehicle`. In our calculator, we always show new and used EV incentives together, so not having the duplicate will actually make for a better user experience. We can (i.e. I can) add `used_electric_vehicle` when we (I) let incentives have multiple items.

In any case, could you reflect the new/used situation in the description? E.g. "$50 rebate for purchasing a new or used EV".

data/PA/incentives.json
```diff
@@ -1089,5 +1089,67 @@
...
+    "program": "pa_pECO_driverRebates",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 50
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "$50 rebate for purchasing a plug-in EV."
```
oyamauchi, line 1152: I'm not sure about the word "plug-in" here. Generally I think pure battery EVs aren't called "plug-in EVs", and the word is used only with plug-in _hybrid_ EVs. If "plug-in EV" is how PECO phrases it, though, I'm ok with it.

src/data/programs/pa_programs.ts
```diff
@@ -80,6 +80,14 @@ export const PA_PROGRAMS = {
       en: 'https://www.peco.com/ways-to-save/for-your-home/rebates-discounts/heating-cooling-rebates',
     },
   },
+  pa_pECO_driverRebates: {
+    name: {
+      en: 'Driver Rebates',
```
oyamauchi, line 85: Add the authority to the program names?

data/PA/incentives.json
rpartridge: their website doesn't specify and there's no way to get more details without logging into a PECO account, so I just assumed it was new to avoid misleading anyone

data/PA/incentives.json
rpartridge: yeah, that's what their website says so I think it's okay 🤷‍♀️

PR #431, opened 2024-04-11
Let /v1/utilities serve utilities for all states

oyamauchi:
## Description

We want to show real utility options to all users, not just in
launched/beta states. We now have the data, so we can remove the check.

https://app.asana.com/0/1206661332626418/1206843501113809

## Test Plan

`yarn test` and updated tests/fixtures. Run queries for some states
that we haven't touched yet.

### DIFF
--- a/test/data/utilities.test.ts
+++ b/test/data/utilities.test.ts
@@ -15,13 +15,17 @@ test('all authorities.json utilities are in DB', async t => {
     .map(Object.keys)
     .flat();
 
+  const query = await database.all<{ utility_id: string; ct: number }[]>(
+    'SELECT utility_id, COUNT(1) AS ct FROM zip_to_utility GROUP BY utility_id',
+  );
+  const utilityCounts = Object.fromEntries(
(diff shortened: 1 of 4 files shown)


test/data/utilities.test.ts
```diff
@@ -15,13 +15,17 @@ test('all authorities.json utilities are in DB', async t => {
     .map(Object.keys)
     .flat();
 
+  const query = await database.all<{ utility_id: string; ct: number }[]>(
+    'SELECT utility_id, COUNT(1) AS ct FROM zip_to_utility GROUP BY utility_id',
+  );
+  const utilityCounts = Object.fromEntries(
+    query!.map(({ utility_id, ct }) => [utility_id, ct]),
+  );
+
```
oyamauchi, line 24: This change isn't strictly necessary but this test had gotten super slow with all the states' utilities added. This speeds it up by doing one DB query instead of thousands

RandomEtc (approved):
Very excited for this. I think it will clear up some of the hiccups people have using the calculator's inline utility selector where we haven't launched state incentives yet.

Approach all looks sound to me.

PR #435, opened 2024-04-15
Fix Maine utilities

oyamauchi:
## Description

A lot of zips seem to be missing from Maine in the Energy Star
dataset, so we're going to end up falling back to listing all the
utilities in many cases.

The list of utilities now pretty much matches this:
https://www.maine.gov/mpuc/regulated-utilities/electricity/delivery-rates

...although see the comment about Versant Power.

## Test Plan

`yarn test`


scripts/lib/utility-data-overrides.ts
```diff
@@ -180,7 +180,15 @@ export const OVERRIDES = new Map<string | number, string>([
   [15263, 'Potomac Edison'],
 
   // ME
+  // Versant has two districts, Bangor Hydro and Maine Public. They have
```
rpartridge, line 183: will this be confusing to users if they don't realize Bangor hydro/Maine Public and Versant are the same? (or maybe that's obvious if you're a customer?) I looked up Bangor hydro and Versant came up so those seem to be the same, but Maine Public Utilities does seem to be its own thing https://www.maine.gov/mpuc/home

rpartridge (approved):
one question but not a blocker

scripts/lib/utility-data-overrides.ts
oyamauchi: That website is the state PUC. If you go to https://www.versantpower.com, it pops up a modal to make you choose the Bangor Hydro or Maine Public district:

Screenshot 2024-04-16 at 10 16 24 PM

So it seems to me that customers would know the name Versant (although it is new within the last few years, that's consistently their customer-facing brand) and also which Versant district they're in.

oyamauchi: Also I inspected the HTML of this form and it does actually have a mapping of zips to districts, right there in the `<select>` tag, so we can use that if we ever need to.

PR #439, opened 2024-04-17
Combine Duquesne weatherization updates

rpartridge:
We got some last minute feedback from Duquesne that they'd like to combine these 5 weatherization incentives ($400 each) into 1 incentive (Up to $2,000) before the public launch on Monday. They confirmed these rebates can be stacked and that they prefer this representation.

Open to ideas on the wording here for the description - this seems okay to me but the wording is a little weird to make sure it's under the 150 character limit.

Unfortunately this means we'll need to get the Spanish translation redone - I'll make a ticket for that but it may not be done by Monday, which I think is fine


oyamauchi (approved):
Wording seems fine to me, and combining these makes the user experience better.

PR #440, opened 2024-04-17
Integrate new AMI data

oyamauchi:
## Description

This is a pretty sweeping change; here's the tour.

- Minor stuff

  - All the old AMI tables are gone from sqlite. state_mfi.json is
    also gone. The `zips` table is still there for zip->zcta and
    zip->state mapping.

  - The new tables are imported with `CREATE TABLE STRICT`, which
    engages SQLite's basic type checking.

- Big stuff

  - This data includes all five populated US territories, so those
    residents can now at least look up federal incentives.

  - I decoupled location lookup from AMI lookup. There's now a
    function that turns location as given in the API (`zip` or
    `address`) into a unified `ResolvedLocation` object that contains
    everything that downstream uses need: state, city, county, ZCTA,
    and optionally census tract.

  - The AMI result is returned as a struct with computed 80% and 150%
    numbers (adjusted to household size using the formula), plus a
    boolean determination of 30C eligibility (since it depends only on
    location, not on the claimant's income).

  - The interaction with the data is vastly simplified: we just pluck
    the right number out of one DB row.

- Tests

  - I had to change several tests because of the new AMI thresholds,
    and the improved 30C calculation. For the most part I left the
    test inputs as-is and updated the output, but in some cases I
    adjusted the income to be in the same AMI bracket as before.

  - Unit tests for the AMI/30C calculation.

https://app.asana.com/0/1206661332626418/1205273704262088

### Notable differences

Unfortunately, we were overestimating 30C eligibility in many
cases. The calculation code had a bug where it was checking whether
the _highest_ poverty rate across all possible tracts was over 20%,
whereas it should have been checking the _lowest_ poverty rate. This
led to false positives in urban ZIPs with a mix of low-poverty and
high-poverty tracts.

### Followups

- Switch to identifying counties (in `authorities.json`) with FIPS
  codes. Our `zips` table and Geocodio have different conventions for
  county names, but FIPS codes are consistent.

- Add special-casing for utilities in the territories, to bring them
  closer to parity with states + DC.

## Test Plan

Tests pass. Manual testing of the API with both `zip` and `address`
params.


src/lib/location.ts
```diff
@@ -0,0 +1,79 @@
...
+ * match it found. (Really it should be "precision", not "accuracy".) These
+ * types are imprecise enough, in practice, that we can't trust the geocoded
+ * census tract.
+ */
+const GEOCODIO_LOW_PRECISION = new Set([
+  'street_center',
+  'place',
+  'county',
+  'state',
+]);
```
oyamauchi, line 23: This is an editorial decision. The other `accuracy_type` values seem to be precise enough that they're not going to possibly span multiple census tracts. See: https://www.geocod.io/docs/#accuracy-score

RandomEtc (approved):
Looks good. Bummer about the EV calculation, glad you caught that.

I am approving / reviewing mostly based on the changes to tests. Where the change in expected result is material, this appears to be correct. Where the change is not material, that's reassuring that the calculations still work end to end and the changes aren't _that_ vast.

This one is probably worth an update the the API users and an update to the changelog in the API docs site. I can do the former, can you do the latter after you cut a release?

RandomEtc:
Re:
> This data includes all five populated US territories, so those residents can now at least look up federal incentives.

That's a great side effect! My understanding is that territories do have rebate allocations based on https://www.energy.gov/save/rebates so that sounds right. However they might not pay Federal income taxes based on https://www.irs.gov/individuals/international-taxpayers/individuals-living-or-working-in-us-possessions - it's probably worth thinking this through but initially I would vote to hide federal tax credits from zip codes outside the 50 states + DC.

PR #450, opened 2024-04-25
Add multi-valued "items" field

oyamauchi:
## Description

This is the beginning of the process of adding more granular item
types for heat pumps and weatherization. Individual incentives will
now pertain to multiple `Item`s.

The process will be:

1. Land and deploy this
2. Land and deploy rewiringamerica/embed.rewiringamerica.org#156
3. Land and deploy a PEP PR which is yet to come, similar to the embed one above (read from new field if available; fall back to existing one if not)
4. Update the backend to stop including the old single-valued field in responses.
5. Remove the support for the old field from the embed and PEP code
6. Start updating the underlying data, this getting rid of the "synthesize `items`" bit from this PR.

I decided to ditch the human-friendly `name`. It was a holdover from
v0, in which it was used to populate the table UIs of the old IRA and
embed calculators. It's too frontend-coupled to make total sense in
v1, and neither PEP nor the new embed calculator use it, so it's gone.

Throughout this I don't expect to proactively inform any external API
clients of the changes, and certainly not to block on them. I believe
anyone we've given v1 to has been warned that v1 will break without
notice.

## Test Plan

`yarn test` with updated fixtures. Tested the embed frontend from the
above PR against this state running locally.

### DIFF
--- a/scripts/update-fixtures.sh
+++ b/scripts/update-fixtures.sh
@@ -112,6 +112,16 @@ zip=60304\
 &authority=il-state-of-illinois" \
  | jq . > test/fixtures/il-60304-state-utility-lowincome.json
 
+ curl \
+  "http://localhost:3000/api/v1/calculator?\
+zip=60202\
+&owner_status=homeowner\
+&household_income=10000\
+&tax_filing=single\
+&household_size=2\
+&authority_types=city" \
+ | jq . > test/fixtures/v1-il-60202-city-lowincome.json
+
 # TODO: Remove beta states argument when MI is fully launched.
 curl \
   "http://localhost:3000/api/v1/calculator\
@@ -124,6 +134,19 @@ curl \
 &authority_types=utility\
 &utility=mi-dte" \
   | jq . > test/fixtures/v1-mi-48103-state-utility-lowincome.json
+
+# TODO: Remove beta states argument when MI is fully launched.
+curl \
+  "http://localhost:3000/api/v1/calculator\
+?zip=48825\
+&include_beta_states=true\
+&owner_status=homeowner\
+&household_income=10000\
+&tax_filing=single\
+&household_size=2\
+&authority_types=utility\
+&utility=mi-lansing-board-of-water-and-light" \
+  | jq . > test/fixtures/v1-mi-48825-city-lowincome.json
 
 curl \
   "http://localhost:3000/api/v1/calculator\
--- a/src/schemas/v1/incentive.ts
+++ b/src/schemas/v1/incentive.ts
@@ -16,6 +16,7 @@ export const API_INCENTIVE_SCHEMA = {
     'program',
     'program_url',
     'item',
+    'items',
     'amount',
     'owner_status',
   ],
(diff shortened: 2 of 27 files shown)


scripts/update-fixtures.sh
```diff
@@ -112,6 +112,16 @@ zip=60304\
 &authority=il-state-of-illinois" \
  | jq . > test/fixtures/il-60304-state-utility-lowincome.json
 
+ curl \
```
matelau, line 115: I didn't realize we have to add to this script as well! I've wrongly assumed the script was aware of the fixtures and updated them accordingly opps. TIL

src/schemas/v1/incentive.ts
```diff
@@ -58,6 +59,13 @@ export const API_INCENTIVE_SCHEMA = {
       required: ['type'],
       additionalProperties: false,
     },
+    items: {
+      type: 'array',
```
matelau, line 63: I'm having a hard time imagining how this will be applied to the situation. I'm familiar with how multiple heat pump incentives appear to be duplicates in our embed and similarly with weatherization incentives. 

Do you have an illustrative example of how this will be used? Are you thinking of grouping similar items in the client side code in embed? As of know, it looks like we are just adding some data redundancies wit hteh type property, but I'm sure I'm just missing something.

scripts/update-fixtures.sh
oyamauchi: You're right though that this is not how one would reasonably expect it to work! These are basically snapshot tests and should work as such. Unfortunately I don't think tapjs supports that concept very well; yet another reason to switch to Jest?

src/schemas/v1/incentive.ts
oyamauchi: I wrote up my thinking on this a few weeks ago here: https://www.notion.so/rewiringamerica/Granular-items-or-sub-items-274423aff1cd42f880367f13510500ae

The tl;dr is that I do anticipate some amount of grouping logic in the frontend, for incentives with multiple items. E.g. if an incentive has `items: [ductless_heat_pump, ducted_heat_pump]`, there would be some logic that recognizes that all the items are types of heat pump and would use the headline `$X off a heat pump`.

I'm not sure what you mean by the type property -- if you mean the `type` property of the `item` property, that's because I'm going to remove `item` altogether once this is rolled out and the embed and PEP are switched over to use `item`.

scripts/update-fixtures.sh
matelau: I'm not sure jest does either out of the box either. I've used something like this in the past https://github.com/philschatz/fetch-vcr that you can configure to "record" the response once and leverage it later. I think it has built in functionality to replace the recorded versions. So, it could possibly be worth exploring.

matelau (approved):

src/schemas/v1/incentive.ts
matelau: I must have missed this. I just checked it out though. Also, thanks for the TLDR.

PR #451, opened 2024-04-29
Identify counties by FIPS code, not by name

oyamauchi:
## Description

Our zips data identifies counties just by their name, e.g. `Suffolk`.
Geocodio identifies them like `Suffolk County`. This would cause
problems for county-level incentives, which need the location's county
to match the county name in `authorities.json`.

It's not trivial to adapt one format into the other (the second word
isn't always "County"; you can't just chop off the last word of the
Geocodio value; etc.) so instead, let's identify counties using a
consistent non-human-friendly identifier.

One other little advantage of the FIPS code is that the first five
digits of a tract ID are the county's FIPS code. We're not making use
of this fact yet, but it may come in handy.

Two concurrent minor changes:

- I removed the county name from the `location` object returned in the
  API, since we're not passing the county name around anymore. We
  could add it back in, but neither the embed nor PEP uses it anyway,
  so I'm thinking that can wait until it's needed.

- I made it so that the `authorities` object in the API response only
  contains names and logos, not the other bits that specify the
  authority's geographic scope. Those aren't part of the public API,
  and I'd rather not let them become so. (Embed and PEP are not using
  them.)

I used this document to manually convert existing county names (it
doesn't have territories, but I just got those from the tract IDs):
https://transition.fcc.gov/oet/info/maps/census/fips/fips.txt

## Test Plan

`yarn test`

### DIFF
--- a/data/NV/authorities.json
+++ b/data/NV/authorities.json
@@ -47,84 +47,84 @@
   "county": {
     "nv-washoe-county": {
       "name": "Washoe County",
-      "county": "Washoe"
+      "county_fips": "32031"
     },
     "nv-churchill-county": {
       "name": "Churchill County",
-      "county": "Churchill"
+      "county_fips": "32001"
     },
     "nv-douglas-county": {
       "name": "Douglas County",
-      "county": "Douglas"
+      "county_fips": "32005"
     },
     "nv-lyon-county": {
       "name": "Lyon County",
-      "county": "Lyon"
+      "county_fips": "32019"
     },
     "nv-storey-county": {
       "name": "Storey County",
-      "county": "Storey"
+      "county_fips": "32029"
     },
     "nv-elko-county": {
       "name": "Elko County",
-      "county": "Elko"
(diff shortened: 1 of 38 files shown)


data/NV/authorities.json
```diff
@@ -47,84 +47,84 @@
   "county": {
     "nv-washoe-county": {
       "name": "Washoe County",
-      "county": "Washoe"
+      "county_fips": "32031"
```
matelau, line 50: Does this fips need to be set by developers or is the `generate-misc-state-data.ts` script taking care of this? 

Is any documentation necessary to inform developers on how to pull this value if it isn't automated?

matelau (approved):
Changes LGTM, but I did have 1 process question.

data/NV/authorities.json
oyamauchi: It needs to be set by engineers. (Setting the `county` key was also done manually.) I'll add something to the readme

PR #452, opened 2024-04-29
Adds Colorado Energy Office Heat Pump Tax Credit Incentives

matelau:
Adds Colorado Energy Office Heat Pump Tax Credit Incentives
Resource: https://energyoffice.colorado.gov/hptc
Asana Ticket: https://app.asana.com/0/1206661332626418/1207142221464012/f

### DIFF
--- a/data/CO/incentives.json
+++ b/data/CO/incentives.json
@@ -7561,8 +7561,8 @@
       "homeowner"
     ],
     "short_description": {
-      "en": "$499.95 discount for ENERGY STAR & AHRI certified system. Available through registered contractors.",
-      "es": "Descuento de $499.95 para sistemas certificados por ENERGY STAR y AHRI. Disponible a través de contratistas registrados."
+      "en": "At least $499.95 discount for ENERGY STAR & AHRI certified system. Available through registered contractors.",
+      "es": "Al menos $499.95 de descuento para sistemas certificados por ENERGY STAR y AHRI. Disponible a través de contratistas registrados."
     },
     "start_date": "2024-01-01",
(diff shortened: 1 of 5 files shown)


data/CO/incentives.json
```diff
@@ -7544,6 +7544,75 @@
...
+    "program": "co_coloradoEnergyOffice_coloradoHeatPumpTaxCredits",
+    "amount": {
+      "type": "dollar_amount",
+      "number": 999.9
+    },
+    "owner_status": [
+      "homeowner"
+    ],
+    "short_description": {
+      "en": "$999.90 discount for ENERGY STAR & AHRI certified system. Available through registered contractors.",
```
matelau, line 7587: This value looks weird, but is the minimum a contractor is required to give the purchaser. see: https://energyoffice.colorado.gov/hptc

oyamauchi (approved):
They're _at least_ that amount, right? Like the tax credit goes to the contractor, and those dollar amounts are the minimum they have to pass through to you. But they _can_ do more. If that understanding is correct, it's probably worth reflecting that in the descriptions, with "at least $999.90" etc.?

matelau:
> They're _at least_ that amount, right? Like the tax credit goes to the contractor, and those dollar amounts are the minimum they have to pass through to you. But they _can_ do more. If that understanding is correct, it's probably worth reflecting that in the descriptions, with "at least $999.90" etc.?

@oyamauchi I'm cool with that! I originally had up to the max amount and then erred on the conservative side. I like your suggestion though. I'll update this.

PR #456, opened 2024-05-01
Add more items to the schema

oyamauchi:
## Description

This does not actually change any of the incentive data; just adds new
granular items to the schema.

I think this will be the best place to discuss the editorial matter of
exactly which items to add. Some of the decisions I made:

- I added `other_heat_pump` as a catchall for heat pump-like devices
  that don't fall under `ducted_heat_pump`, `ductless_heat_pump`,
  `central_air_conditioner`, `air_to_water_heat_pump`, or
  `geothermal_heating_installation`. I don't know if such a thing
  really exists, or if it does whether we care to include it in the API?

  - Note that my plan is to deprecate
    `heat_pump_air_conditioner_heater`, rather than use it as the catchall.

- I also added (a) `other_insulation` and (b) `other_weatherization`.
  Respectively, these are for (a) insulation that isn't covered by one
  of the many granular categories I added; and (b) any weatherization
  measure that is not insulation and is not covered by one of the
  granular non-insulation measures. Similar to above, my plan is to
  deprecate `weatherization` rather than use it as the catchall.

- I added `{new,used}_plugin_hybrid_vehicle`. We've seen several
  incentives for thesespecifically, and I think it's worth splitting
  them out from `{new,used}_electric_vehicle`. The latter will become
  pure-EV only.

On the side, I made it so that human-friendly names and URLs are only
required for the items that appear in the IRA incentives (since
they're only needed in API v0).

### DIFF
--- a/src/data/types/items.ts
+++ b/src/data/types/items.ts
@@ -4,7 +4,18 @@
  */
 
 export const ITEMS_SCHEMA = {
+  air_sealing: { type: 'string' },
+  air_to_water_heat_pump: { type: 'string' },
+  attic_or_roof_insulation: { type: 'string' },
+  basement_insulation: { type: 'string' },
   battery_storage_installation: { type: 'string' },
+  central_air_conditioner: { type: 'string' },
+  crawlspace_insulation: { type: 'string' },
+  door_replacement: { type: 'string' },
+  duct_replacement: { type: 'string' },
+  duct_sealing: { type: 'string' },
+  ducted_heat_pump: { type: 'string' },
+  ductless_heat_pump: { type: 'string' },
   ebike: { type: 'string' },
   efficiency_rebates: { type: 'string' },
   electric_outdoor_equipment: { type: 'string' },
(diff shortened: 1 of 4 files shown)


src/data/types/items.ts
```diff
@@ -13,22 +24,51 @@ export const ITEMS_SCHEMA = {
...
-  heat_pump_air_conditioner_heater: { type: 'string' },
+  heat_pump_air_conditioner_heater: { type: 'string' }, // TODO remove
   heat_pump_clothes_dryer: { type: 'string' },
   heat_pump_water_heater: { type: 'string' },
   new_electric_vehicle: { type: 'string' },
+  new_plugin_hybrid_vehicle: { type: 'string' },
   non_heat_pump_clothes_dryer: { type: 'string' },
   non_heat_pump_water_heater: { type: 'string' },
   other: { type: 'string' },
+  other_heat_pump: { type: 'string' },
```
oyamauchi, line 39: On second thought maybe this should be `other_hvac`? The only thing I can remember seeing in incentives that isn't covered by the other categories is window units, which I guess is technically a type of heat pump. I can't immediately think of a non-covered HVAC item that we would want to show that isn't some kind of heat pump.

src/data/types/items.ts
```diff
@@ -13,22 +24,51 @@ export const ITEMS_SCHEMA = {
...
   rooftop_solar_installation: { type: 'string' },
   smart_thermostat: { type: 'string' },
   used_electric_vehicle: { type: 'string' },
-  weatherization: { type: 'string' },
+  used_plugin_hybrid_vehicle: { type: 'string' },
+  wall_insulation: { type: 'string' },
+  weatherization: { type: 'string' }, // TODO remove
   whole_house_fan: { type: 'string' },
+  window_replacement: { type: 'string' },
 } as const;
```
2WheelCoder, line 50: I think we're missing `electric_lawn_equipment`?

2WheelCoder: For the sake of completeness I'll make an argument that `water_efficiency` should be represented, but I get how that's not really our thing. Energy auditors are taught to consider water efficiency as energy efficiency because it a utility cost and it takes energy to clean water, build water infrastructure and transport water. I also think a climate impacted world, and the same audience who is thinking about electrification, will find it relevant. Do I think we should include it, though? I have no idea.

oyamauchi: We have `electric_outdoor_equipment`; slightly more general, but I feel like it's not worth making distinctions finer than that.

I get the argument for water efficiency, but I think that's a separate question from how to split up existing categories. It seems pretty clearly not a subset of any of the existing items.

src/data/types/items.ts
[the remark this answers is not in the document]
oyamauchi: Yeah I think I'm going to stick with `other_heat_pump`, and if we start wanting to add incentives for non-heat-pump HVAC equipment (other than `whole_house_fan` and `evaporative_cooler` which we already have), we can add enum values for those later.

src/data/types/items.ts
2WheelCoder: Ah, I missed `electric_outdoor_equipment`. LGTM!

2WheelCoder (approved):

PR #459, opened 2024-05-03
Store multiple items per incentive in JSON

oyamauchi:
## Description

The API is already serving an `items` array; now make the JSON data
store an `items` array and parse multiple values from the
corresponding spreadsheet column.

None of the spreadsheets have multiple values in that column yet, so
the data changes are purely mechanical at this point, not substantive.

I did the data changes by running the script in most cases (any state
for which the spreadsheet health check passes) and doing a find/replace
in others (CT, RI, NY, VA, IRA).

One thing to note is that I had to update any spreadsheet with electric
lawn equipment incentives, because the spreadsheet value for that
category contains commas, which breaks the comma-separation of multiple
values that this field is now doing. I changed the canonical value to
`Electric outdoor equipment`, and updated the incentives and enum value
sheets.

## Test Plan

`yarn test`.


oyamauchi:
After this, I need to wait until embed and PEP can handle all the new values (that's rewiringamerica/embed.rewiringamerica.org#158 for the embed), then I can start setting new values for the field. That will be a long slog.

test/data/schemas.test.ts
```diff
@@ -226,18 +225,6 @@ test('state incentives JSON files match schemas', async tap => {
   });
 });
 
-test('All items used in IRA incentives have urls available', async tap => {
```
matelau, line 229: Is this test case no longer valid?

matelau (approved):
One question, no blockers 👍🏼

test/data/schemas.test.ts
oyamauchi: This invariant started being enforced by JSON Schema validation in #456 , so the test was already redundant. Now `incentive.item` is no longer a thing, so rather than update the test I deleted it.

PR #462, opened 2024-05-07
States Endpoint

2WheelCoder:


src/data/types/states.ts
```diff
@@ -72,10 +72,3 @@ export const LAUNCHED_STATES: string[] = [
   'VA',
   'VT',
 ];
-
-export const isStateIncluded = (
-  stateId: string,
-  includeBeta: boolean,
-): boolean =>
-  LAUNCHED_STATES.includes(stateId) ||
-  (includeBeta && BETA_STATES.includes(stateId));
```
2WheelCoder, line 81: I thought I'd move this to `src/lib/states.ts` with the rest of the states related functions, but let me know if that's not the right thinking.

oyamauchi (approved):
Looks good!

PR #463, opened 2024-05-09
Remove PRPA incentives as they duplicate the utilities

matelau:
Remove PRPA incentives as they duplicate the utilities.

See discussion here: https://app.asana.com/0/1206770021434525/1207089549000817/f

https://www.prpa.org/about-prpa/
```
Platte River Power Authority is a not-for-profit public power utility that generates reliable, financially sustainable and environmentally responsible electricity for our owner communities of Estes Park, Fort Collins, Longmont and Loveland.
```

These incentives are provided via[ Efficiency Works](https://efficiencyworks.org/) but our including PRPA incentives in the API made it seem as though the Efficiency Works incentives were duplicated. 
image (2)


oyamauchi (approved):
This accomplishes the goal, but I'm wondering: why not delete the versions that are specific to the utilities (e.g. CO-49, CO-50), and keep the ones you've deleted here?

For one thing, all the utility-specific versions are copies of each other, which is a maintenance headache. For another, I think having a single record for each incentive more closely reflects reality -- it seems you don't actually interact with your utility to get these rebates, and the utilities don't have their own info about them.

The programs for the utility-specific versions (`co_lovelandWaterAndPower_efficiencyWorks`, etc.) all have the same name (`Efficiency Works`) and URL, so we wouldn't lose any differentiation there. The geo group handles the utility-based eligibility.

Approving to unblock you in case I'm missing some good reason not to do that. But I would rather go with the approach I suggested if possible.

matelau:
> This accomplishes the goal, but I'm wondering: why not delete the versions that are specific to the utilities (e.g. CO-49, CO-50), and keep the ones you've deleted here?
> 
> For one thing, all the utility-specific versions are copies of each other, which is a maintenance headache. For another, I think having a single record for each incentive more closely reflects reality -- it seems you don't actually interact with your utility to get these rebates, and the utilities don't have their own info about them.
> 
> The programs for the utility-specific versions (`co_lovelandWaterAndPower_efficiencyWorks`, etc.) all have the same name (`Efficiency Works`) and URL, so we wouldn't lose any differentiation there. The geo group handles the utility-based eligibility.
> 
> Approving to unblock you in case I'm missing some good reason not to do that. But I would rather go with the approach I suggested if possible.

You know I don't know why I didn't take that approach. I agree with your take. Thanks! I'll get this updated.

oyamauchi (approved):
This looks good. Did you set the "omit from API" flag in the spreadsheet? I know the sheet-to-JSON scripts don't work right now and you must have made these changes manually, but setting the flag will prevent these changes from being undone next time someone runs the script.

matelau:
> This looks good. Did you set the "omit from API" flag in the spreadsheet? I know the sheet-to-JSON scripts don't work right now and you must have made these changes manually, but setting the flag will prevent these changes from being undone next time someone runs the script.

Yep, that has been done!

PR #464, opened 2024-05-09
Unlock calculation for territories

oyamauchi:
## Description

We can do this now that we have AMI data and 30C eligibility data for
the territories. This further required:

- Making the solar system cost estimation optional

- Making the estimated savings lookup fall back to the `US` value if
  it's not known for the user's state.

## Test Plan

Tested old and new embed frontends against local API server, with zip
00930 (PR).  The old one correctly shows total tax savings of $0 for
low and high incomes. The rooftop solar credit is shown as "30%"
instead of an estimated amount.

The tax credits still show up as eligible, which I think is not ideal,
but is OK -- territory residents are technically eligible for the
credits, and may benefit from them if they have non-territory-sourced
income or some other circumstance that means they do pay nonzero
federal income tax. (And this is the same thing you see for a low
income in the states.)

### DIFF
--- a/src/lib/tax-brackets.ts
+++ b/src/lib/tax-brackets.ts
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { FilingStatus, TAX_BRACKETS } from '../data/tax_brackets';
+import { TERRITORIES } from '../data/types/states';
 import { TaxEstimate } from '../data/types/tax';
 import { roundCents } from './rounding';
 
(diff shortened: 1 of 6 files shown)


src/lib/tax-brackets.ts
```diff
@@ -32,9 +33,18 @@ export function estimateStateTaxAmount(
 }
 
 export function estimateFederalTaxAmount(
+  state: string,
   filing_status: FilingStatus,
   household_income: number,
 ): TaxEstimate {
+  // Bona fide residents of the territories don't pay federal income tax on
```
matelau, line 40: Do we need to express this assumption someway via the UI or API docs?

src/routes/v0.ts
```diff
@@ -199,7 +199,8 @@ export default async function (
 
           // 2) Add annual savings from pregenerated model
           estimated_annual_savings:
-            IRA_STATE_SAVINGS[location.state].estimated_savings_heat_pump_ev,
+            IRA_STATE_SAVINGS[location.state]?.estimated_savings_heat_pump_ev ??
+            IRA_STATE_SAVINGS['US'].estimated_savings_heat_pump_ev,
```
matelau, line 203: Nit: Is this fallback data? Should we store this as CONST somewhere to make that more expressive? IE: `const DEFAULT_HEAT_PUMP_EV_SAVINGS = ...`?

matelau (approved):
I had a couple questions, but otherwise looks good to me!

PR #465, opened 2024-05-10
Remove the singular item field from v1

oyamauchi:
## Description

This is the last piece to unlock updating the incentives to have
multiple items.

Embed frontend is already updated to not read this field.

Don't merge until rewiringamerica/consumer#1138 is deployed.

## Test Plan

`yarn test`.


matelau (approved):
👏🏼

PR #466, opened 2024-05-13
Add CORS plugin for Fastify, in non-production

oyamauchi:
## Description

I keep having to apply this locally when testing the embed frontend
against local API changes, I figure it'll probably be generally
useful.

The plugin is disabled by means of `NODE_ENV` in the Docker build,
because for now the Docker image is only used in prod, and in prod we
have Zuplo handling CORS for us.

## Test Plan

Run `yarn dev` and run the embed frontend with the API host set to
`http://localhost:3000`; make sure calculations succeed, and the CORS
requests are visible in web inspector and in the server log.

Build and run the Docker image; run the embed frontend against it and
make sure the CORS preflight requests fail.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,5 +38,6 @@ COPY data data
 COPY locales locales
 
 # go!
+ENV NODE_ENV=production
 ENV PORT=8080
 CMD yarn start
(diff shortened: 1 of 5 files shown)


RandomEtc (approved):
Looks good to me, all makes sense.

I think Zuplo wants you to have parity in local dev as well, and their solution to that is their CLI that lets you run your gateway locally. Might be worth a look if you haven't already ruled it out: https://zuplo.com/docs/articles/local-development

matelau (approved):

oyamauchi:
That makes sense, but using the Zuplo gateway isn't an option for people running this code who aren't us, so I think we'll still want this anyway.

PR #469, opened 2024-05-14
Add lint rule to enforce braces on control flow statements

oyamauchi:
## Description

I'm personally opposed to `if (condition) doStuff()`; I think it's
best to always have braces in those situations to avoid silly and
potentially quite consequential mistake. This adds an eslint rule to
enforce that, and fixes all existing instances.

## Test Plan

`yarn lint`, `yarn test`


matelau (approved):
👍🏼

PR #470, opened 2024-05-15
Run spreadsheet health check by default

oyamauchi:
## Description

I would love it if we didn't add any more spreadsheets to the
registry, but in case we do, I want them to be used with the script,
and passing the health check, from the start, so make it the default
to run the health check.

## Test Plan

`yarn check-spreadsheet-health` has the same result as before. (There
are a few states failing because of new translations.)


matelau (approved):
> I would love it if we didn't add any more spreadsheets to the
registry, but in case we do, I want them to be used with the script,
and passing the health check, from the start, so make it the default
to run the health check.

soon!

PR #471, opened 2024-05-15
Exclude IRA rebates from v1

oyamauchi:
## Description

The embed is already filtering these out and replacing them with its
own content. I'd say PEP shouldn't be showing them at all, since
they're not actionable.

We do still need to keep them in the v0 API, to avoid disrupting the
legacy embed and v0 API clients.

The boolean flag is inelegant, but it will be removed along with the
v0 API.

## Test Plan

`yarn test` with new/updated tests.


src/lib/incentives-calculation.ts
```diff
@@ -76,6 +77,14 @@ function calculateFederalIncentivesAndSavings(
       continue;
     }
 
+    if (
```
matelau, line 80: Nit: Could we put the computation of the boolean in a named function to make it easier to understand? ie: `if (shouldExcludeV0(incentive.payment_methods[0]){ continue }` or something with a better name lol.

matelau (approved):

PR #477, opened 2024-05-16
Refreshing expired CT data

mariabajzek:
Updated the source spreadsheet as well and created a named revision for it.

For CT I checked things manually since there's not a lot of data and since I wanted to see what this process would be like. It wasn't bad but seems like it would be pretty awful for e.g. CO so I need to look at the automation work that was done for this and maybe pick that back up.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -6,5 +6,6 @@ module.exports = {
   root: true,
   rules: {
     eqeqeq: ['error', 'always'],
+    curly: ['error', 'all'],
   },
 };
--- a/data/AZ/incentives.json
+++ b/data/AZ/incentives.json
@@ -30,7 +30,7 @@
       "rebate"
     ],
     "items": [
-      "heat_pump_air_conditioner_heater"
+      "ducted_heat_pump"
     ],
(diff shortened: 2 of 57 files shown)


mariabajzek:
Hi @oyamauchi , it's been a minute since I've worked on this code - do you know how changing dates in this one file could have caused all these tests to fail? I did also notice failures locally so I thought maybe I need to go through and install stuff again (/ install new things that have been added since I left in March).

(no rush to get back to me about it immediately, I can take another look at this tomorrow if the problem is not obvious to you)

oyamauchi:
It looks like the failing tests are just because the fixture file needs to be updated. Make sure the server is running locally and then run `scripts/update-fixtures.sh`.

mariabajzek:
Hmm, very strange... I was seeing a lot of database-related issues (missing tables, nonexistent authorities, etc) and the update-fixtures script replaced all output with 500 errors, but I ran it AGAIN and I think things look ok now.

oyamauchi (approved):
Ah, then I think the problem was needing to run `yarn build` to recreate the SQLite database. I've meant to set up Husky in this repo and add a post-checkout hook to run that automatically if necessary, but never gotten around to it.

PR #478, opened 2024-05-16
Split items in IRA incentives

oyamauchi:
## Description

This splits up the combined items in the IRA incentives, allowing me
to get rid of `heat_pump_air_conditioner_heater` and `weatherization`
from the main items enum.

To keep the v0 API unchanged, I've added a new `v0_item` field to the
IRA incentive data. The values of that enum are no longer a subset of
the main item enum, because the two named above are still in
there. The field is allowed to be null, to exclude the incentive from
v0 altogether; I've done this with the new incentive for the energy
audit portion of the 25C tax credit. (`energy_audit` doesn't exist in
v0, and it doesn't make sense to have this part of the credit as a
separate `weatherization` incentive.)

## Test Plan

`yarn test`. In particular make sure the only change in the v0 fixture
is the changed short description.


src/data/types/items.ts
```diff
@@ -44,31 +43,38 @@ export const ITEMS_SCHEMA = {
...
-  'new_electric_vehicle',
-  'rooftop_solar_installation',
-  'used_electric_vehicle',
-  'weatherization',
-] as const satisfies readonly Item[];
+/**
+ * These items are for the legacy v0 API, and only used with IRA incentives.
+ *
+ * Note that this contains two items that don't appear above: weatherization and
+ * heat_pump_air_conditioner_heater. In v1 (and state and utility incentives),
```
matelau, line 56: 👍🏼

locales/en.json
```diff
@@ -17,28 +17,19 @@
   },
   "items": {
     "battery_storage_installation": "Battery Storage Installation",
-    "ebike": "E-Bike",
```
matelau, line 20: Did we intend to drop so many of these? I couldn't see how these modifications were related to the vo_item changes so I just wanted to raise.

matelau (approved):

locales/en.json
oyamauchi: This is because I set the `items` and `urls` fields of the locale files to contain IRA items only. (The strings are only used in v0, and v0 only contains IRA incentives, and those don't use any of the removed items.)

PR #485, opened 2024-05-21
Support per-county low income thresholds

oyamauchi:
## Description

This has been hanging over our heads since Vermont, but now NY's HEAR
program has them, so it's time to actually do it.

I added a `type` field to each threshold definition, which can have
the values `hhsize` or `county-hhsize`, indicating the sequence of
keys to be used with the `thresholds` structure. Counties are
identified by FIPS code, same as in the authorities files.

To demonstrate this working, I updated the county-specific Vermont
incentives (it's a lot less data than the NY ones, which I'll do
separately).

I've also switched the low-income-thresholds JSON-Schema-to-TS logic
away from ajv and to `json-schema-to-ts`. The latter is what I prefer
to use for this because it's what Fastify uses to turn endpoint
schemas into types, and I think it's worthwhile to have everything use
the same solution for that (because every schema-to-TS thing has
subtle differences). I've been converting things gradually. Anyway,
this resulted most notably in having to remove the `required:
[1,2,...]` constraint on the household-size-thresholds struct, so I
added a unit test for that.

## Test Plan

New test for county-specific Vermont thresholds (same income is LI in
one county but not in another). Other tests pass.

### DIFF
--- a/test/data/low_income_thresholds.test.ts
+++ b/test/data/low_income_thresholds.test.ts
@@ -1,5 +1,5 @@
 import { test } from 'tap';
-import { LOW_INCOME_THRESHOLDS_BY_AUTHORITY } from '../../src/data/low_income_thresholds';
+import { LOW_INCOME_THRESHOLDS_BY_STATE } from '../../src/data/low_income_thresholds';
 import {
   STATE_INCENTIVES_BY_STATE,
   StateIncentive,
@@ -18,7 +18,7 @@ test('low-income thresholds are equivalent in JSON config and incentives', async
   const map = computeIdToIncentiveMap(
     Object.values(STATE_INCENTIVES_BY_STATE).flat(),
   );
-  Object.entries(LOW_INCOME_THRESHOLDS_BY_AUTHORITY).forEach(
+  Object.entries(LOW_INCOME_THRESHOLDS_BY_STATE).forEach(
     ([, stateThresholds]) => {
       for (const [identifier, thresholds] of Object.entries(stateThresholds)) {
         for (const incentiveId of thresholds.incentives) {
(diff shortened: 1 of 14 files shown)


test/data/low_income_thresholds.test.ts
```diff
@@ -52,3 +52,22 @@ test('low-income thresholds are equivalent in JSON config and incentives', async
     }
   }
 });
+
+test('low-income thresholds have HH sizes 1-8', async t => {
+  const hasRequiredKeys = (obj: Record<string, unknown>) => {
+    const keys = Object.keys(obj);
+    return [1, 2, 3, 4, 5, 6, 7, 8].every(num => keys.includes(num.toString()));
```
matelau, line 59: Question: I noticed some thresholds go up to ten. Do we ever use those values in the API? I'm mostly curious about this for data modeling purposes in the admin app.

matelau (approved):
I had one question for my own understanding, but otherwise this looks good. I think maybe we need to include some documentation though to outline how to add low income thresholds for a specific county.

test/data/low_income_thresholds.test.ts
oyamauchi: Values above 8 aren't ever used because the `household_size` request param is limited to >= 1 and <= 8. But we put the higher numbers in the data when the original source had them, just in case.

Extra info:

I think the request param is limited to 8 because the official AMI numbers (which were relevant for the IRA incentives) were only published up to HH size = 8.

In the case of AMI for IRA incentives, there's a simple formula to extend the thresholds to bigger HH sizes. But I don't think there generally is such a formula for all the various state and utility thresholds. And a lot of those only give thresholds up to 8, so I think we're generally stuck with 8 as the upper limit.

oyamauchi (commented):
I forgot to mention: I'm pretty sure this scheme can accommodate the `filing_status`, `agi_max_limit` and `agi_min_limit` filtering as well, which would let us get rid of those fields. Possibly `ami_qualification` too?

PR #488, opened 2024-05-22
Add NY's EmPower+; remove other NY incentives

oyamauchi:
## Description

We need to add this program to the calculator very soon, but the other
NY incentives we've gathered haven't been looked at in several months,
and they haven't been through the most up-to-date QA process at all.
Rather than rushing to QA them in time for EmPower's launch, I'm just
taking them out entirely.

They're coded as `efficiency_rebates` for expediency. This `item`
value is displayed by the frontend as "an energy efficiency retrofit",
which seems to be how EmPower+ is presented: it's a whole package deal.

The income thresholds came in the form of Javascript on the NYSERDA
eligibility page, so I wrote a quick script to populate them in the
low-income-thresholds file; that's included here. (There's one set of
thresholds per county in NY, so I didn't want to do it all manually.)

## Test Plan

Run the frontend against local API server; make sure it shows up
reasonably. (I posted an example here:
https://rewiringameri-g3x1100.slack.com/archives/C0711MM7K8R/p1718665502271849 )

For the script: run it and look at the resulting thresholds changes.
Spot-check a few of the FIPS codes to make sure it matches what the
NYSERDA webpage says if you use their calculator tool.


scripts/ny-empower-income-limits.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return result;
+}
+
+async function main() {
+  const db = await open({
+    filename: path.join(__dirname, '../incentives-api.db'),
+    driver: sqlite3.Database,
+  });
+  console.log(db);
```
matelau, line 43: Did you want to keep this log or was it for debugging purposes?

scripts/ny-empower-income-limits.ts
```diff
@@ -0,0 +1,166 @@
...
+  for (const [countyName, raw] of data.entries()) {
+    if (countyName === 'NY') {
+      continue;
+    }
+
+    const fips = await getCountyFips(db, countyName);
+    console.log(countyName, fips);
+    moderateThresholds[fips] = convertToObject(raw);
+  }
+  console.log(moderateThresholds);
```
matelau, line 72: Similarly, did we mean to leave this log? If so, it could benefit from some additional context ie: `console.log("Generated threshold - ny-empower-moderate:", moderateThresholds);`

matelau (approved):
Nothing major here from me, I did have one question in do we need to document the usage of this?
It seems like you generated the data now and the script is now superfluous, but it is good to keep around in case the data needs future modifications. 

Also, how do we keep track of the state of the NY spreadsheet? seems like executing of the typical conversion scripts could undo the incentive data done here.

oyamauchi:
The script is meant to be long-lived; I expect the thresholds will change regularly (every year, I think) and if they're still encoded on the page as a JS variable, the script will still be useful.

The scripts currently error out if you run them on NY spreadsheet (it's pretty old and has never been in script-ready shape), so I'm not worried about someone running the scripts and wiping out these changes. As extra assurance I can put a comment in `spreadsheet-registry.ts` where the NY entry would go.

PR #490, opened 2024-05-24
Replace update-fixtures.sh with env var in test itself

oyamauchi:
## Description

It's kinda silly to have to define every test case in two places like
this. This is a strategy I've used before: have an env var that
controls whether the test does the "snapshot == live" assertion, or
just overwrites the snapshot with the live value.

The workflow for adding a new test case is to write the
`validateResponse` call in v1.test.ts, including the query parameters
and the path for the new snapshot file, then run `yarn
update-snapshots` to generate the initial snapshot. Manually check to
make sure it's as expected.

For updating an existing test case, just change the query parameters
in the test case (if necessary) and run `yarn update-snapshots`. Check
to make sure the changes are as expected.

There's no way to selectively update individual snapshots; I can add
that if it proves necessary, but it might not.

While I'm at it, update the naming of these things to "snapshot"
rather than "fixture", as that's what it is.

## Test Plan

Make sure all the fixture/snapshot files are unchanged.

Go through the above workflows for adding a new test case and updating
an existing one. `yarn test` still passes.


.prettierrc
```diff
@@ -9,7 +9,12 @@
   ],
   "overrides": [
     {
-      "files": ["data/*.json", "schemas/**/*.json", "test/fixtures/*.json"],
+      "files": [
+        "data/*.json",
+        "schemas/**/*.json",
+        "test/fixtures/*.json",
```
matelau, line 15: is fixtures still necessary?

CONTRIBUTING.md
```diff
@@ -81,7 +81,7 @@ We use Prettier and ESLint; both are run for every PR and are required to report
 
 The codebase has fairly comprehensive test coverage, and we expect all changes to be well-covered. There are:
 
-- End-to-end tests of the API endpoints, in `test/routes`. These compare live output against fixtures in `test/fixtures`.
-  - There's a script to update the fixtures to match the output of a local instance of the server, in `scripts/update-fixtures.sh`. If you add a new fixture, make sure to add it to the script.
+- End-to-end tests of the API endpoints, in `test/routes`. These compare live output against snapshots in `test/snapshots`.
+  - To update the snapshots from the live output, run `yarn update-snapshots`. You can also use this command to create the initial snapshot for a new test.
```
matelau, line 85: 👍🏼

matelau (approved):

.prettierrc
oyamauchi: Yeah, there's some stuff in there that is actually test fixtures (sample incentive relationships, etc.)

PR #494, opened 2024-05-28
Don't return ineligible incentives from v1

oyamauchi:
## Description

The fact that ineligible incentives were returned from `/calculator`
is a holdover from v0 that no longer makes sense. In v0, it was done
support showing ineligible IRA incentives grayed-out. But in the new
calculator, we're not doing that. Also, as a matter of API design,
it's conceptually unclean: it makes more sense if the endpoint just
returns a set of incentives that fulfill the criteria passed in.

Both embed and PEP clients are immediately filtering out ineligible
incentives anyway, so this doesn't affect them at all.

The next step will be to remove the `eligible` field from the API.

## Test Plan

`yarn test`. Look at the snapshot diffs and make sure all the removed
incentives have `eligible: false`. Grep the snapshots directory for
`"eligible": false` and make sure the only hit is in the v0 snapshot.


src/routes/v1.ts
```diff
@@ -94,7 +94,10 @@ export default async function (
         );
         const translated = {
           ...result,
-          incentives: transformIncentives(result.incentives, language),
+          incentives: transformIncentives(
```
matelau, line 97: nit:  the filtering happening nested here could be easy to overlook. Do you think it would be valuable to pull this up at at all? For example...
```javascript

const result = ... ;
const filteredIncentives = result.incentives.filter(i => i.eligible);
const translated = {...result, incentives: transformIncentives(filteredIncentives, langage) ...
```

It can sometimes be hard to grok when you have multiple nested operations.

matelau (approved):

PR #495, opened 2024-05-28
[WIP] Update VT incentive relationships

matelau:
This nested prerequisite was giving me issues in incentive admin when trying to seed the data with the existing data. I wanted to check if the relationship as defined is equivalent to the pre-existing nested relationship.


data/VT/incentive_relationships.json
```diff
@@ -14,10 +14,9 @@
     "VT-86": "VT-84",
     "VT-89": "VT-87",
     "VT-93": "VT-91",
+    "VT-94": {"anyOf": ["VT-75", "VT-77", "VT-79", "VT-81"]},
     "VT-96": {
-      "allOf": [
-        "VT-94",
-        { "anyOf": ["VT-75", "VT-77", "VT-79", "VT-81"] }]
+      "allOf": ["VT-94"]
```
matelau, line 19: Another alternative here could be something like, which would also work downstream in incentive admin... 
```json
    "VT-96": {
      "allOf": ["VT-94"],
      "anyOf": ["VT-75", "VT-77", "VT-79", "VT-81"]
    }
```

matelau: But, if we go this route we would also need to update the schema to allow for more than oneOf the expected definitions. 
image

matelau: Thinking about this more and I don't think either solution works. I'm going to consider if I can create a recursive solution in Incentive Admin to follow the nesting. Open to ideas though. 
image

oyamauchi:
So VT-96 is all cleaned up now, but the schema simplification may still be relevant? Or maybe not, since ideally we won't be bringing in significant amounts of new incentive data before the admin tool is online?

matelau:
> So VT-96 is all cleaned up now, but the schema simplification may still be relevant? Or maybe not, since ideally we won't be bringing in significant amounts of new incentive data before the admin tool is online?

I hope not 🤞🏼! I'll close this PR and we can re-visit if necessary.

PR #499, opened 2024-06-04
Document all the fields

oyamauchi:
## Description

Added description strings for almost all the request and response
fields of all endpoints.

I didn't describe all the possible `items` enum values. That's going
to be very long, and I think is best left for its own separate
page. I'll add a link to that page from the doc string, once it's
ready.

There are some other OpenAPI hygiene changes like adding summaries for
endpoints (which makes them show up better in doc generators), and
adding operation IDs.

I snuck in two substantive changes:

- The same schema that represents authorities in the backend JSON data
  was being used to represent them in the API. However, the JSON has
  more fields than we want to expose in the API, so create a new
  schema for authorities that excludes them.

- Remove the `include_beta_states` parameter from the utilities
  endpoint, since it's not read anymore. I already deployed a change
  to the embed to stop passing it, and made sure PEP doesn't use it.

## Test Plan

Use a work-in-progress script I have that merges the spec generated
from `/spec.json` into Zuplo's config file. The result is deployed
here:

https://api-main-4ba10f3.d2.zuplo.dev/docs

That doesn't show the descriptions of response fields and subfields (a
known shortcoming of Zuplo). However, you can use third-party doc
generator tools with the OpenAPI spec. Here are two examples:

- https://redocly.github.io/redoc/
- https://elements-demo.stoplight.io/

In both of those, enter this URL in the "try it" box to see the docs:

https://api-main-4ba10f3.d2.zuplo.dev/spec.json

### DIFF
--- a/src/schemas/v1/calculator-endpoint.ts
+++ b/src/schemas/v1/calculator-endpoint.ts
@@ -14,20 +14,20 @@ const API_CALCULATOR_REQUEST_SCHEMA = {
   properties: {
     zip: {
       type: 'string',
-      description:
-        'Your zip code helps us estimate the amount of discounts and tax credits you qualify for by finding representative census tracts in your area.',
+      description: `Find incentives that may be available in this ZIP code. \
+Exactly one of this and "address" is required.`,
       maxLength: 5,
       minLength: 5,
     },
     address: {
       type: 'string',
-      description:
-        "Your address can determine the precise census tract you're in that determines the correct amount of discounts and tax credits you qualify for.",
+      description: `Find incentives that may be available at this address. \
+Exactly one of this and "zip" is required.`,
     },
     authority_types: {
       type: 'array',
-      description:
-        'Which types of authority to fetch incentives for: "federal", "state", or "utility".',
+      description: `Find incentives offered by these types of authorities. \
+If absent, incentives from all types of authorities will be considered.`,
       items: {
         type: 'string',
         enum: Object.values(AuthorityType),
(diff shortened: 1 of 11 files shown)


src/schemas/v1/incentive.ts
```diff
@@ -82,12 +139,20 @@ export const API_INCENTIVE_SCHEMA = {
...
     end_date: {
       type: 'string',
-    },
-    special_note: {
-      type: 'string',
+      description: `The date when the incentive stopped, or will stop, being \
+available. Format is the same as for \`start_date\`.`,
     },
     ami_qualification: {
       type: 'string',
```
oyamauchi, line 158: I haven't bothered with these four fields because I'm planning to drop them soon. (All of these concepts can be absorbed into the low-income-thresholds concept, and I think it might be better to expose low-income thresholds in the API the same way we express them in the backend, if at all.)

src/schemas/v1/calculator-endpoint.ts
```diff
@@ -37,14 +37,14 @@ const API_CALCULATOR_REQUEST_SCHEMA = {
...
+      description: `The ID of your utility company, as returned from \
+\`/api/v1/utilities\`. Required if authority_types includes "utility". If \
+absent, no incentives offered by utilities will be returned.`,
     },
     items: {
       type: 'array',
-      description:
-        'Which types of product or project to fetch incentives for. If absent, include all products/projects.',
+      description: `Which types of product or service to fetch incentives for. \
+If absent, include incentives for all products and services.`,
```
matelau, line 47: Do we need to expose a list of valid items that can be used here somewhere?

src/schemas/v1/incentive.ts
```diff
@@ -74,6 +129,8 @@ export const API_INCENTIVE_SCHEMA = {
     },
     owner_status: {
       type: 'array',
+      description: `Whether the consumer must be a homeowner or renter (or \
+either) to claim this incentive.`,
```
matelau, line 133: Do we present homeowner or renter as either in the api or is it returned as `both?

src/schemas/v1/incentive.ts
matelau: Do you think we should mark these fields as deprecated then?

matelau (approved):
LGTM, I commented on some minor things but NBD.

src/schemas/v1/incentive.ts
RandomEtc: We use multiple values:

```
    "owner_status": [
      "homeowner",
      "renter"
    ],
```

src/schemas/v1/calculator-endpoint.ts
oyamauchi: Zuplo generates a list of values; it looks like this:

Screenshot 2024-06-05 at 6 02 12 PM

Like I said in the commit message: because there are so many items, it would get way too long to include all of them here, and I'm planning to explain them on a separate page. I'll add a link to the separate page here (and in the `items` response field's docs) when it's ready.

PR #500, opened 2024-06-07
Replace the "agi" fields with a new type of low-income threshold

oyamauchi:
## Description

The `agi_max_limit` and `agi_min_limit` fields are fundamentally about
income qualification.  As such, I think it makes sense to subsume it
into the wider concept of low-income thresholds. That lets us get rid
of the awkward duplication of incentives that's required to express
different AGI thresholds for different filing statuses.

The `filing_status` field has no use other than mapping filing status
to different AGI thresholds, so it's gone too. I'm leaving it as a
"collected" field from spreadsheets so that I don't have to go update
all the spreadsheets, but it's no longer in the actual incentives
JSON.

In theory `ami_qualification` should get the same treatment, but I
think it may not actually be long for the world -- the only incentives
that use it are the generic IRA rebates, and those will be gone after
v0 is shut down.

I haven't updated the VT spreadsheet yet, in case we decide to delay
or drop this change.

I've made sure the embed frontend does not read these fields, and I've
got rewiringamerica/consumer#1198 to remove them from PEP. (They
weren't being used.)

## Test Plan

`yarn test`. Had to update some tests and fixtures to fit the new regime.


matelau (commented):
These changes LGTM! Thanks for jumping in on these. I think you're right it'll make the work downstream easier and it doesn't seem like it'll be a huge modification.  

Should we add some additional filing-status tests in: `test/data/low_income_thresholds.test.ts` ?

oyamauchi:
Added a check that `min < max`, and that they're both > 0. (Checking that a min/max pair is present for all filing statuses is done by the JSON Schema validation.)

matelau (approved):

oyamauchi:
PEP change is deployed, so this is good to go.

PR #505, opened 2024-06-14
Change most instances of STATES_PLUS_DC to STATES_AND_TERRITORIES; clean up states endpoint spec

oyamauchi:
## Description

At this point we should be treating states, DC, and territories as
equivalent by default. There are a couple of exceptions:

- The territories are listed separately because the tax calculation
  needs to special-case them as zero federal income tax.

- States plus DC are listed separately because two data files
  (ira_state_savings and solar_prices) only have entries for states
  and DC. The former is only relevant for v0, and the latter is only
  used to estimate solar system cost (and the result is not surfaced
  anywhere in our current frontends), so it doesn't seem worth the
  effort to try to fill in the equivalent data for the territories.
  Those are the only uses of `STATES_PLUS_DC` remaining.

While I'm at it, change the `/v1/states` endpoint spec to not list out
all the states explicitly. Instead, explain in the description what
the keys are, and just document the `{ status: "..." }` struct once.

## Test Plan

`yarn test`.

### DIFF
--- a/src/data/authorities.ts
+++ b/src/data/authorities.ts
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import { FromSchema } from 'json-schema-to-ts';
 import { API_IMAGE_SCHEMA } from '../schemas/v1/image';
-import { STATES_PLUS_DC } from './types/states';
+import { STATES_AND_TERRITORIES } from './types/states';
 
 /**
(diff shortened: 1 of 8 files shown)


matelau (approved):
LGTM!

PR #511, opened 2024-06-20
Update VT incentives

benjf:
Update the incentives for VT.

Asana link: https://app.asana.com/0/1207456219977120/1203684030594289

Testing: `yarn test`

### DIFF
--- a/data/VT/incentives.json
+++ b/data/VT/incentives.json
@@ -218,7 +218,8 @@
       "rebate"
     ],
     "items": [
-      "ducted_heat_pump"
+      "ducted_heat_pump",
+      "ductless_heat_pump"
     ],
     "program": "vt_greenMountainPower_greenMountainPowerDuctedHeatPumpRebate",
     "amount": {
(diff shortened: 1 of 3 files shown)


data/VT/incentives.json
```diff
@@ -218,7 +218,8 @@
       "rebate"
     ],
     "items": [
-      "ducted_heat_pump"
+      "ducted_heat_pump",
+      "ductless_heat_pump"
     ],
     "program": "vt_greenMountainPower_greenMountainPowerDuctedHeatPumpRebate",
```
oyamauchi, line 224: This program's name still says "Ducted Heat Pump Rebate". (It's defined in `src/data/programs/vt_programs.ts`.) Could you:

- rename the program key, here and in vt_programs.ts, to remove the word "ducted";
- remove the word "Ducted" from the user-visible name in vt_programs.ts?

The `*_programs.ts` files were originally generated from spreadsheets by script (`scripts/generate-misc-state-data.ts`), but that was basically a one-time thing when the state was first added, and now the VT spreadsheet has drifted enough that the script currently errors out. So I think hand-editing to fix this is the way forward. (Hopefully the spreadsheets' days are numbered anyway.)

oyamauchi (changes requested):
This is right, but there's one thing left to update.

matelau (approved):

oyamauchi (approved):

PR #522, opened 2024-07-01
Override "City of Eugene" (OR) with municipal utility name

oyamauchi:
They're pretty consistently branded as "EWEB" (https://www.eweb.org/)
so going with that rather than spelling it out.


scripts/lib/utility-data-overrides.ts
```diff
@@ -293,6 +293,8 @@ export const OVERRIDES = new Map<string | number, string>([
   [14109, 'Oregon Trail Electric Cooperative'],
   [18917, "Tillamook People's Utility District"],
   [1736, 'Blachly-Lane Electric Cooperative'],
+  [6022, 'EWEB'], // Eugene Water & Electric Board
+  [430373, 'EWEB'],
```
rpartridge, line 297: it's been a while since I've worked on this... remind me why we need line 297 as well? are there 2 records for this utility in the energy star data set?

rpartridge (approved):

scripts/lib/utility-data-overrides.ts
oyamauchi: Yeah there are two different utility IDs for EWEB, for some reason. One of them (430373) is only associated with what seems to be a non-real zip code (00069); I don't know what's going on with that, but I figure there's no harm in overriding both.

PR #523, opened 2024-07-01
Update WI incentives

oyamauchi:
## Description

Just ran the script. Changes match the description in the task:
https://app.asana.com/0/1204990513426536/1207633035019166

The snapshot diff is unfortunately confusing because the ordering has
changed (due to the payment methods of a bunch of incentives
changing).

## Test Plan

`yarn test`


hblumberg (approved):
🦡

PR #535, opened 2024-07-17
Refreshing data for IL

mariabajzek:
I also created a saved revision of the IL spreadsheet to mark the updates that I made after checking the incentives.


data/IL/incentives.json
```diff
@@ -229,16 +232,16 @@
...
+      "maximum": 1800
     },
     "owner_status": [
       "homeowner"
     ],
     "short_description": {
-      "en": "Up to $1,500 rebate for All-Electric Cold Climate Air Source Heat Pump. Minimum efficiency of 15.2 SEER2 & 7.6 HSPF2.",
-      "es": "Un reembolso de hasta $1,500 para una bomba de calor de fuente de aire de clima frío totalmente eléctrica. Eficiencia mínima de 15.2 SEER2 y 7.6 HSPF2."
+      "en": "Up to $1,800 rebate for All-Electric Cold Climate Air Source Heat Pump. Minimum efficiency of 15.2 SEER2 & 7.6 HSPF2.",
+      "es": "Un reembolso de hasta $1,800 para una bomba de calor de fuente de aire de clima frío totalmente eléctrica. Eficiencia mínima de 15.2 SEER2 y 7.6 HSPF2."
```
veekas, line 244: would the idiomatic way to write the Spanish dollar value here be `$1.800` instead? (I know the previous revision also used a comma)

veekas (approved):
Owen should probably still take a look but this looks good to me

data/IL/incentives.json
oyamauchi: We've been using US English-style numbers in Spanish everywhere in the API and UI. I don't remember who, but someone familiar with Spanish said that would be OK since our audience is only people in the US, and US Spanish is fine with "," thousand separator and "." decimal point.

mariabajzek: Sounds good, thanks for confirming! I'll merge it then.

PR #549, opened 2024-07-29
New low-income threshold type for AMI percentage; use in WI

oyamauchi:
## Description

Focus on Energy's income-qualified incentives are now based on AMI,
rather than per-state thresholds like last year. Fortunately, we
already have up-to-date AMI data at hand, so just plumb it in to the
low-income calculation.

I confirmed that the numbers on their site match the numbers in our
AMI database in a few randomly selected counties.

## Test Plan

`yarn test` with new snapshot test (which fails on the old income thresholds).

### DIFF
--- a/data/low_income_thresholds.json
+++ b/data/low_income_thresholds.json
@@ -1864,7 +1864,9 @@
         "WI-30"
       ],
       "source_url": "https://focusonenergy.com/income-qualified",
-      "percentage": 80
+      "thresholds": {
+        "percentage": 80
+      }
     }
   }
 }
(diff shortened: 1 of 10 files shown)


oyamauchi:
This implicates the incentives admin code by having a new threshold type; I'm working on a change to adapt to this.

matelau (approved):
Changes look good to me! I created this ticket to also address these changes in the incentive admin. https://app.asana.com/0/1207304265252553/1207912619420873/f

veekas (approved):
looks great, thanks!

PR #556, opened 2024-08-21
Fix Denver healthy homes titles

cjohnsbrown:
## Description

The titles on the embed for the Denver healthy homes incentives was set to "$0 off an electric/induction stove". I updated the sheet to a 100% discount to instead show "100% of cost of an electric/induction stove".

## Test

* `yarn test`
* Tested changes with local embed instance


veekas (approved):
good catch thanks!

PR #562, opened 2024-08-27
Add APS Incentives

veekas:
Description

- adds APS as a utility and maps it to the zip codes in their service area (see paragraph below)
- adds APS Home Performance Programs incentives

In order to do this, I had to create a sheet to add additional utilities that aren't in the Energy Star database. I plan to connect with the Data Science team to see if there's a way to upstream this step so we can avoid having to manually manage this.

Test

`yarn build`
`yarn test`


cjohnsbrown (approved):

scripts/lib/additional-utility-zips.xlsx
RandomEtc, line 1: For future, not blocking, I think that we might be able to determine the EIA ID for APS and use that rather than a synthetic ID in this file (or similar files in future). It's not load bearing today because we don't propagate that ID. But if we upstream this or start working with patches, or we start working with the open data shapefiles for utility coverage, it might be helpful.

data/AZ/programs.json
```diff
@@ -294,5 +294,13 @@
     "url": {
       "en": "https://www.garkaneenergy.com/energy-efficient-rebates"
     }
+  },
+  "az_arizonaPublicService_homePerformanceProgram": {
+    "name": {
+      "en": "Home Performance Program"
```
RandomEtc, line 300: Again not blocking, but because these program names currently get displayed standalone in the calculator I think for clarity it might be good to put APS or Arizona Public Service in here? I think we are not consistent about this today, maybe worth discussing as a team (the alternative would be to display authority somehow in the calculator, wider Design question?)

RandomEtc:
(Defer to Cameron for actual code review, but excited to land this!)

scripts/lib/additional-utility-zips.xlsx
veekas: I updated it to the EIA Utility ID for APS: 803

data/AZ/programs.json
veekas: That makes sense. I added it to our core sync agenda for the next meeting

PR #568, opened 2024-09-09
Add optional "description" fields to a few types

oyamauchi:
## Description

Since none of the existing prereq relationships uses recursive
anyOf/allOf relationships, and it's super hard to support in a
relational format, I removed support for that. Now prereq
relationships are just `allOf: [list of strings]` or `anyOf: [list of
strings]` or just a single string.

Adding the field to exclusion relationships meant I had to add an
option to expand the list of IDs to an object, with an `ids` key (the
list of IDs) and the description.

## Test Plan

`yarn test`

Import from my local incentive admin with changes applied to export
descriptions in these positions, and make sure tests still pass here.

### DIFF
--- a/src/data/geo_groups.ts
+++ b/src/data/geo_groups.ts
@@ -5,6 +5,7 @@ import { STATES_AND_TERRITORIES } from './types/states';
 const GEO_GROUP_SCHEMA = {
   type: 'object',
   properties: {
+    description: { type: 'string' },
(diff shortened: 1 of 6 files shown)


oyamauchi:
Once the admin change is landed, and exclusions are all exported as `{ids: [...]}`, I'll remove support for the plain list-of-strings option here, to keep things clean.

PR #570, opened 2024-09-12
Move authority and authority_type onto program instead of incentive

oyamauchi:
## Description

Since incentive-admin stores these things as "incentive has a program,
program has an authority", which reflects the underlying reality. It
causes some difficulty that this codebase stores them as "incentive
has a program, incentive has an authority".

This makes it so that programs refer to authorities, and incentives
don't.

- Application-code changes are actually pretty minimal.

- This uncovered a bunch of programs that aren't used anymore; I
  removed those by hand.

- The CT data (very old; predates spreadsheet-JSON scripts) had some
  programs that overlapped multiple utilities, so I split them up by hand.

## Test Plan

`yarn test` with a bunch of updated tests.

With a corresponding admin change, successfully imported and exported.

The spreadsheet-to-JSON scripts are updated enough to pass unit tests,
but I admit I did not test them with real spreadsheets. I'm hopeful
that we won't need to use them much longer.

### DIFF
--- a/data/AZ/incentives.json
+++ b/data/AZ/incentives.json
@@ -1,8 +1,6 @@
 [
   {
     "id": "AZ-2",
-    "authority_type": "utility",
-    "authority": "az-mohave-electric-cooperative",
     "payment_methods": [
       "rebate"
     ],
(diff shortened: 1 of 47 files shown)


cjohnsbrown (approved):
LGTM

PR #578, opened 2024-09-19
Add Maine incentives to API

veekas:
- [Spreadsheet](https://docs.google.com/spreadsheets/d/1vMk9tCokKU3JqyyCVjm7wmZ-4jOKmSCQgDFt6_wHHhk/edit?gid=995688950#gid=995688950)


veekas:
@oyamauchi any theories about why my tests are failing? It seems to be related to the tests not being able to find a matching incentive for given IDs, but the incentives are definitely there...

example error: `me-low-income reference nonexistent incentive ME-1`

oyamauchi:
You have to add code to read the file to the bottom of `src/data/state_incentives.ts`. (Honestly this is kind of silly --- it's all boilerplate --- but we ought to fix it separately from this PR.)

veekas:
@oyamauchi thanks! I forgot about that file. Tests are passing now

cjohnsbrown (approved):

data/low_income_thresholds.json
```diff
@@ -1039,6 +1039,64 @@
...
+          70000
+        ],
+        "married_filing_separately": [
+          0,
+          70000
+        ],
+        "qualifying_widower_with_dependent_child": [
+          0,
+          70000
+        ]
```
oyamauchi, line 1096: I know the source PDF just says the limit for "joint filers" is 100k, but I think that would apply to this status as well. It's meant for people who used to file jointly with their spouse until their spouse died. (Also known as Qualifying Surviving Spouse or QSS.)

data/ME/incentive_relationships.json
```diff
@@ -0,0 +1,8 @@
+{
+  "exclusions": {
+    "ME-1": ["ME-2"],
+    "ME-6": ["ME-7"],
+    "ME-8": ["ME-9"],
```
oyamauchi, line 5: I think these two should be the other way around? ME-6 and ME-8 are the moderate-income versions; instead it should be the low-income version excludes the moderate-income version. (I.e. if we show you the low-income incentive, you shouldn't also see the moderate-income incentive even though your income is necessarily below the moderate thresholds.)

oyamauchi (approved):
Approving to unblock, but the comments are substantive

data/ME/incentive_relationships.json
veekas: Yup that's right, thanks!

PR #580, opened 2024-09-19
Add updated RI HEAR Incentives

veekas:


cjohnsbrown (approved):

data/RI/incentives.json
```diff
@@ -683,5 +683,225 @@
...
+      "en": "Up to $1,250 rebate for electric wiring. Available for moderate-income residents."
+    }
+  },
+  {
+    "id": "RI-53",
+    "payment_methods": [
+      "rebate"
+    ],
+    "items": [
+      "other"
```
oyamauchi, line 869: Should list the items here: dryer and stove.

It's very awkward to express this in the spreadsheet because Sheets doesn't let you do a multiple-selection dropdown, but the spreadsheet-JSON scripts will actually parse comma-separated values in the Technology column. You have to turn off the data validation on that column in the sheet, manually put in the correct Technology values but comma-separated, then turn validation back on. (E.g. I did this in row 3.)

If you don't want to do all that nonsense, I think it's OK to just do it in JSON manually, and don't bother doing it in the spreadsheet.

oyamauchi (changes requested):
All the HEAR incentives talk about being for low and moderate income households but don't have `low_income` fields?

veekas:
Addressed your comments, @oyamauchi. I'm going to merge this in, but if you see something else, I'll do it as a fast follow today. Trying not to let Comms down re: getting this out today

data/RI/incentive_relationships.json
```diff
@@ -9,6 +9,21 @@
     ],
     "RI-43": [
       "RI-13"
+    ],
+    "RI-45": [
+      "RI-46"
+    ],
+    "RI-47": [
+      "RI-48"
+    ],
```
oyamauchi, line 18: These are interesting. Because the pairs (45-46 and 47-48) are identical in all the filtering fields including low income, the excluded ones (46 and 48) will just never appear for anyone.

I think the thing to do is to consolidate each pair into a single incentive, and note the difference caused by the electrical-upgrade aspect in the description. Like the combination of 45 and 46 could be:

```
Rebate of 50% of cost for a Level 2 EV charger, up to $350, or up to $700 if an electrical upgrade is needed.
```

(And have 700 as the `maximum` in the amount)

What do you think?

data/RI/incentives.json
```diff
@@ -858,15 +863,18 @@
...
   {
     "id": "RI-53",
     "payment_methods": [
       "rebate"
     ],
     "items": [
-      "other"
+      "heat_pump_clothes_dryer",
+      "electric_stove",
+      "electric_stove"
```
oyamauchi, line 877: Don't need stove twice

data/RI/incentives.json
```diff
@@ -880,15 +888,18 @@
...
   {
     "id": "RI-54",
     "payment_methods": [
       "rebate"
     ],
     "items": [
-      "other"
+      "heat_pump_clothes_dryer",
+      "electric_stove",
+      "electric_stove"
```
oyamauchi, line 902: same

oyamauchi (approved):
Approving to unblock, but one of the comments is substantive

data/RI/incentives.json
veekas: right, I guess the induction and electric stoves both map to the same key here. I'll delete

PR #585, opened 2024-10-01
update Efficiency Maine mobile home heat pump incentive

veekas:
From EM: "One other inconsistency is on the heat pump offers. We do not have a rebate for $2,000 for a ducted system. I think what is trying to be expressed there is that we have a Manufactured (Mobile) Home Initiative that pays the complete installation of a heat pump and removal of the fossil system with a customer co-pay of $2,000, where Efficiency Maine pays the remaining cost ($12,900)"

### DIFF
--- a/data/ME/incentives.json
+++ b/data/ME/incentives.json
@@ -358,14 +358,15 @@
     ],
     "program": "me_efficiencyMaine_efficiencyMaineResidentialHeatPumpIncentives",
     "amount": {
-      "type": "dollar_amount",
-      "number": 2000
+      "type": "percent",
+      "number": 0.85,
+      "maximum": 12900
     },
     "owner_status": [
       "homeowner"
     ],
     "short_description": {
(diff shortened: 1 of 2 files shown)


data/ME/incentives.json
```diff
@@ -358,14 +358,15 @@
...
+      "type": "percent",
+      "number": 0.85,
+      "maximum": 12900
     },
     "owner_status": [
       "homeowner"
     ],
     "short_description": {
-      "en": "$2,000 rebate for ducted heat pump, with option to be distributed in monthly payments. Available for low income residents."
+      "en": "Up to 85% of the cost of ducted heat pump systems for manufactured (mobile) homes. Available for low-income households. $2,000 co-pay."
```
oyamauchi, line 369: Urgh, I wonder if co-pays are something we have to start representing... I think we're OK without it for now, but that may change if we see more incentives like this.

oyamauchi (approved):

data/ME/incentives.json
veekas: yeah I agree. I'll make a ticket to track the concept

PR #590, opened 2024-10-15
Re-export from HERO with consistent key and list ordering

oyamauchi:
## Description

I landed a change in HERO to make JSON exports have object keys in a
consistent order, and also to sort lists whose order doesn't matter.
This should make these diffs much easier to read going forward.

This should be a no-op diff in terms of substance: I imported from API
`main` into my local HERO, then exported from that.

The one thing that's not strictly an ordering change is the way
exclusions are represented. They are now:

```
"AZ-1": {
  "ids": ["AZ-2", "AZ-3"]
}
```

instead of

```
"AZ-1": ["AZ-2", "AZ-3"]
```

The API code already treats these the same.

## Test Plan

`yarn test`


data/MI/programs.json
```diff
@@ -109,22 +109,22 @@
     "authority": "mi-indiana-michigan-power",
     "authority_type": "utility"
   },
-  "mi_lansingBoardOfWaterAndLight_lansingBWLElectrificationPrograms": {
+  "mi_lansingBoardOfWaterAndLight_lansingBWLENERGYSTARApplianceRebates": {
```
oyamauchi, line 112: This churn is just because HERO sorts case-sensitively. That was easier to implement in HERO than case-insensitive sorting would have been, and this churn is a one-time cost.

data/AZ/incentive_relationships.json
```diff
@@ -1,12 +1,44 @@
...
-    "AZ-83": ["AZ-84"],
-    "AZ-85": ["AZ-86"],
-    "AZ-87": ["AZ-88"],
-    "AZ-89": ["AZ-90"],
-    "AZ-91": ["AZ-92"],
-    "AZ-93": ["AZ-94"]
+    "AZ-79": {
+      "ids": [
+        "AZ-80"
+      ]
```
veekas, line 6: more verbose/cluttered at this point, but good to be explicit anyway

veekas (approved):

data/AZ/incentive_relationships.json
oyamauchi: We shouldn't be hand-editing these anymore anyway, and ideally won't even really be looking closely at them.

PR #591, opened 2024-10-17
Export additive changes from HERO

oyamauchi:
## Description

HERO and this codebase are currently out of sync -- both sides have
changes that the other side does not have. This is the first step to
resolving that situation.

I did a full export from HERO, and created this PR from only the parts
that were _added_; these are the updates that were in HERO but not in
JSON/spreadsheets.

With this done, we can once again do non-destructive imports from JSON
to HERO, to bring them fully back in sync and complete the deprecation
of spreadsheets.

There are two minor changes to the validation tests:

- Bumped the English character limit for descriptions up to 250, to
  accommodate a long description for a VT incentive that we were asked
  to put in. We do have a path to reducing this again, though; I've
  made a ticket for it.

- Removed `minItems: 1` from the JSON schema for low-income
  thresholds, because there's now a set of thresholds in MI that has
  no incentives. The check was there in the manually-edited-JSON days
  basically to catch mistakes where you forgot to set the threshold on
  the incentives that used it; that's less of an issue now, so I feel
  OK losing this check.

## Test Plan

`yarn test`.

If I import from prod HERO now, the only changes are removing a bunch
of ME and RI data. That's data that was added via spreadsheets, and
not reflected in HERO.

### DIFF
--- a/test/data/schemas.test.ts
+++ b/test/data/schemas.test.ts
@@ -76,6 +76,9 @@ import {
 } from '../../src/lib/incentive-relationship-calculation';
 import { incentiveRelationshipsContainCycle } from './cycles';
 
+const ENGLISH_DESC_CHAR_LIMIT = 250;
+const SPANISH_DESC_CHAR_LIMIT = 400;
+
 const TESTS = [
   [I_SCHEMA, IRA_INCENTIVES, 'ira_incentives'],
   [ISS_SCHEMA, IRA_STATE_SAVINGS, 'ira_state_savings'],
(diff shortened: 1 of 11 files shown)


test/data/schemas.test.ts
```diff
@@ -76,6 +76,9 @@ import {
 } from '../../src/lib/incentive-relationship-calculation';
 import { incentiveRelationshipsContainCycle } from './cycles';
 
+const ENGLISH_DESC_CHAR_LIMIT = 250;
+const SPANISH_DESC_CHAR_LIMIT = 400;
```
veekas, line 80: 🙏🏽

veekas (approved):

PR #596, opened 2024-10-22
[NOT FOR MERGE] Second-to-last round of spreadsheet imports

oyamauchi:
This is the second-to-last-ever spreadsheet import. The last one will be to pull in all the extra columns. I'm doing these imports separately to make reviewing easier.

This branch contains two separate commits. The first is just exactly what resulted from me running the import script on all the spreadsheets. The second is me making manual changes, mostly to revert things where either the script made irrelevant ordering changes, or where HERO is actually more up to date than the spreadsheet.

The end result is that the changes are:

- Adding a bunch of GA incentives that were never reflected anywhere other than the spreadsheet, and
- Setting a VT incentive's item back to `other`. (VT-74 is a rebate for a _wood_ stove, not an electric stove.)


veekas (approved):

oyamauchi:
This is all imported into HERO now

PR #597, opened 2024-10-22
[NOT FOR MERGE] Data for final HERO import

oyamauchi:
This is where all the new columns are added.

**The new columns should be the only changes**. If you spot any other changes, that is a bug.

No need to actually go verify the content of every single addition -- these columns are for internal use only, not shown in the API.

Also, the code changes are not going to be merged (all the spreadsheet scripts are about to be thrown out anyway) so no need to review them in depth; I included them in this PR just to show what I did to produce the data updates.

I've tested importing this state into my local HERO instance, and it seemed to work fine. The next step, if the changes here look good, is to import into staging HERO.


veekas (approved):
thanks for leaving the script updates in for review. looks good

oyamauchi:
All this is now imported into HERO

PR #602, opened 2024-10-28
Delete spreadsheet scripts

oyamauchi:
## Description

At least in theory, we're never importing from spreadsheets into this
codebase again, so it's time to get rid of all that code.

This also unlocks a bunch of cleanup in the definition of the
StateIncentive type: we no longer need to maintain separate schemas
for "collected" and derived incentives.

There's one piece I've left in place:
`authority-and-program-updater.ts`, which is used by the
`generate-utility-data` script. I have a separate plan to get rid of
that, but I want to keep it runnable for now.

Looking at all this, I have mixed feelings. Of course I'm delighted to
be deleting thousands of lines of code, but I'm a little rueful that
we spent all that time and effort writing them in the first place, and
wrangling all those spreadsheets, when we could have been building
HERO months earlier instead. But hey, it got us through the first wave
of state launches, so it wasn't a waste.

## Test Plan

`yarn test`.


coopr:
Maybe silly question, and surely the wrong place to ask it, but... what is HERO?

mpusto:
@coopr HERO stands for the Home Electrification Resource Organizer. It's currently an internal tool we use to manage all authorities and incentive programs. It's not currently documented or open for public use but you can see it here. https://incentive-admin.rewiringamerica.org

PR #614, opened 2024-11-26
Import zip-to-utility from HERO

oyamauchi:
## Description

I've loaded the contents of zip-to-utility.csv as they are now into
the HERO DB, which will now be the source of truth for it. I ran an
import and that's where the changes to zip-to-utility.csv come from.

This also deletes the `generate-utility-data.ts` script and associated
files. I do anticipate fishing these back out of version control at
some point, for various reasons, but for now I'm deleting them to
avoid confusion.

## Test Plan

Run an import successfully.

Write a throwaway script to verify that the data in the old CSV file
and the new one is identical except for ordering.


veekas (approved):
nice

PR #620, opened 2024-12-06
Add "none" option to gas_utility param; validate param

oyamauchi:
## Description

While implementing the Mass Save logic I realized that there's a
crucial distinction between a user _not having gas service_, and a
user _not specifying who their gas utility is_. (The former case
allows us to show more non-Mass Save utility incentives.)

My solution is to allow passing a special value of `none` to the
`gas_utility` param, meaning you affirmatively do not have gas
service. I'm anticipating that the calculator UI will offer options
like:

- National Grid
- Eversource
- Other
- I don't have gas service

"I don't have gas service" will result in the calculator sending the
special `none` value of the `gas_utility` param, while "Other" will
result in the param being absent.

This also adds logic to validate the parameter.

## Test Plan

`yarn test` with new tests.

Test coverage of this isn't going to be quite comprehensive until we
have the actual Mass Save data in here, which requires a couple more
changes first.

### DIFF
--- a/src/data/authorities.ts
+++ b/src/data/authorities.ts
@@ -98,3 +98,10 @@ export const AUTHORITIES_BY_STATE: AuthoritiesByState = (() => {
   }
   return result;
 })();
+
+/**
+ * A special value for the API to allow positively indicating "I don't have gas
+ * service" (as opposed to your gas utility being unknown or unspecified). This
+ * matters for some eligibility decisions, such as with MA's Mass Save program.
+ */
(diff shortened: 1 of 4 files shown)


coopr:
The choices of "Other" and "I don't have gas service" might be confusing to someone with delivered propane - they might not know which one to pick.

veekas (approved):
looks great!

(Reviewed the front end changes separately that address @coopr's helpful comment by adding a "Delivered propane or fuel oil" option that maps to no gas service on the backend.)

PR #622, opened 2024-12-11
Stop manually loading in each state's incentives

oyamauchi:
## Links

- https://app.asana.com/0/1208668890181682/1208930378081802

## Description

This hasn't made sense since we stopped having programs defined in
Typescript. Just iterate over the states like we do for other types of
state-specific data file.

This will avoid an easily-forgotten piece of busywork that happens
with every new state we add.

## Test Plan

`yarn test`. Manually make a non-conformant change to a state
incentives JSON file and make sure the test fails.


veekas (approved):
nice update

PR #625, opened 2024-12-17
Slightly rework gas utilities API

oyamauchi:
## Description

This will make the API more suitable for general use, while still
supporting the calculator's intended behavior of only showing the gas
utility selector when necessary.

In addition to the recently added `gas_utilities` key in the
`/v1/utilities` response, I added `gas_utility_affects_incentives`.
If we don't have gas utility data for a state at all, both keys are
absent. If we have gas utility data, `gas_utilities` is always
populated and `gas_utility_affects_incentives` is present.

This way, clients who are just interested in which utilities serve the
area get whatever data we have. But the calculator can look at the
boolean flag to see whether it should show the selector.

## Test Plan

Testing is incomplete because we only have gas utility incentives in
MA, which is special-cased in the code anyway. But the new and
existing `/utilities` tests are decent coverage.

### DIFF
--- a/test/routes/v1.test.ts
+++ b/test/routes/v1.test.ts
@@ -904,6 +904,26 @@ const UTILITIES = [
       },
     },
   ],
+  [
+    '02116',
+    {
+      location: { state: 'MA', city: 'Boston', county_fips: '25025' },
+      utilities: {
+        'ma-eversource': {
+          name: 'Eversource',
+        },
+      },
+      gas_utilities: {
+        'ma-national-grid-gas': {
+          name: 'National Grid',
+        },
+        'ma-nstar-gas-company': {
+          name: 'Eversource',
+        },
+      },
+      gas_utility_affects_incentives: true,
+    },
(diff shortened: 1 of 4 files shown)


test/routes/v1.test.ts
```diff
@@ -904,6 +904,26 @@ const UTILITIES = [
...
+      },
+      gas_utilities: {
+        'ma-national-grid-gas': {
+          name: 'National Grid',
+        },
+        'ma-nstar-gas-company': {
+          name: 'Eversource',
+        },
+      },
+      gas_utility_affects_incentives: true,
```
cjohnsbrown, line 924: So the gas utilities will always be determined and returned, and this flag will just be used to let the client decide whether to show them or not?

cjohnsbrown (approved):

test/routes/v1.test.ts
oyamauchi: Right

PR #650, opened 2025-01-24
Add unit test for existence of city authorities in zips table

oyamauchi:
## Description

Ticket: https://app.asana.com/0/1208668890181682/1209141933112058

In `resolveLocation`, we pull a postal city and county FIPS out of the
`zips` table in SQLite. Then those are matched against `city`-type
authorities. This test makes sure that each `city` authority would
match at least one zip, because if it doesn't, that city authority is
effectively dead.

## Test Plan

`yarn test`


oyamauchi:
This is mooted by the geography work I'm doing.

PR #655, opened 2025-01-30
Use static date for snapshot tests

cjohnsbrown:
Filtering out expired incentives from `/calculator` can break some snapshot tests when using the actual current time if an incentive has expired. This change utilizes a static date during testing in order to keep the response values consistent. I set the date to 1/25/2025 since anything older would include incentives that have already been removed from the current snapshot tests.

## Test
- `yarn test`
- Unset the `STATIC_DATE` envar and `yarn test` fails because an incentive expired on 1/27 and will not be included in the response

### DIFF
--- a/src/lib/state-incentives-calculation.ts
+++ b/src/lib/state-incentives-calculation.ts
@@ -137,17 +137,7 @@ export function calculateStateIncentivesAndSavings(
 
     if (item.end_date) {
       const lastValidDay = lastDayOf(item.end_date);
-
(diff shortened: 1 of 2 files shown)


src/lib/state-incentives-calculation.ts
```diff
@@ -381,3 +371,21 @@ function skipBasedOnRequestParams(
 
   return false;
 }
+
+function getCurrentDateTime() {
```
oyamauchi, line 375: I'd call it `getCurrentDate` (it's just a date, not a date+time)

PR #660, opened 2025-02-07
Add loans schema

benjf:


src/data/types/capital-types.ts
```diff
@@ -0,0 +1,159 @@
+import { FromSchema } from 'json-schema-to-ts';
+
+export enum AuthorityType {
```
matelau, line 3: For several of these can't we just import them from our [capital-common](https://github.com/rewiringamerica/capital-common/blob/main/src/types.ts#L4-L13) pkg? 

The implementation seems to be the same and it should help us in maintaining these across the various systems.

src/schemas/v1/loan-programs.ts
```diff
@@ -0,0 +1,138 @@
+import { FromSchema } from 'json-schema-to-ts';
+import { LoanProgramStatus, ProjectType } from '../../data/types/capital-types';
+import { STATES_AND_TERRITORIES } from '../../data/types/states';
+
+const STATES = [...STATES_AND_TERRITORIES, 'NT'];
+
+/**
+ * API response schema for loan programs.
+ */
+export const API_LOAN_PROGRAMS_RESPONSE_SCHEMA = {
```
matelau, line 10: I know you're following the patterns in this project, but are these steps necessary for us given we already have ts and these definitions are transformed to typescript? 

Looking at the code here for [IRA_INCENTIVES](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/cad98a7f3126640efe3d311b0bf7c54230b4be30/src/data/ira_incentives.ts#L127C14-L129) are these definitions only used to type the data that is returned when the file is parsed? 

It may be worth checking with @oyamauchi to see if we already have typescript definitions if we should still define the JSON schema. I could see why it may be useful to continue that practice for maintainability, and visibility, but if it is just a means to an end to get the TS types, we already got them.

src/data/types/capital-types.ts
benjf: Yep, good call!

src/data/types/capital-types.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export const IMAGE_SCHEMA = {
+  $id: 'Image.schema.json',
+  type: 'object',
+  properties: {
+    src: { type: 'string' },
+  },
+  required: ['src'],
+  additionalProperties: false,
+} as const;
```
oyamauchi, line 19: There's already an image schema in `src/schemas/v1/image.ts` (which admittedly should be in `src/data` somewhere instead). We should use that instead of defining a duplicate one.

That other schema has width and height as required properties, so you would need to add those to whatever data source you're going to use for this capital data. (I think it's correct that they're required; it improves rendering on the web if image dimensions are known statically.)

src/schemas/v1/loan-programs.ts
oyamauchi: Zooming out a bit, the JSON Schemas have three purposes in this codebase:

1. To validate the checked-in JSON data files for incentives etc., and to derive Typescript types for that data.
2. To tell Fastify about each API endpoint. ([See here for example](https://github.com/rewiringamerica/api.rewiringamerica.org/blob/cad98a7f3126640efe3d311b0bf7c54230b4be30/src/routes/v1.ts#L73).) Fastify's input and output validation is driven by the JSON Schema, and the types of the "request" and "reply" objects Fastify gives us are derived from the JSON Schema (under the hood, it uses the same `FromSchema` utility seen here).
3. To eventually make their way to our Zuplo config in `api-zuplo`, which drives the public API docs.

So this code here is necessary for 2 and 3.

One possible solution is to pull the JSON Schema, and the `FromSchema`-derived TS types, into common code? Off the top of my head I'm not 100% sure if that'll work, but I think it's the most obvious way to avoid the code duplication you're concerned about here.

src/data/types/capital-types.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Schema for FinancialAuthority.
+ */
+export const FINANCIAL_AUTHORITY_SCHEMA = {
+  $id: 'FinancialAuthority.schema.json',
+  type: 'object',
+  properties: {
+    id: { type: 'number' },
+    authority_type: { type: 'string', enum: Object.values(AuthorityType) },
+    name: { type: 'string' },
+    description: { type: ['string', 'null'] },
```
oyamauchi, line 44: I gotta admit I didn't know `type` could be an array until seeing this. That could have saved me a lot of trouble in the early days :/

src/schemas/v1/loan-programs.ts
benjf: Thanks for the feedback and additional context!

matelau (approved):
✅ But, I think we need to leverage the existing image schema. https://github.com/rewiringamerica/api.rewiringamerica.org/pull/660/files#r1951707413

rserur (approved):

PR #665, opened 2025-02-12
Update tax data to 2025 tax year

oyamauchi:
## Description

Taken from here: https://www.irs.gov/pub/irs-drop/rp-24-40.pdf

These are the brackets that will apply to stuff you do this year, for
the tax return that you'll file in calendar year 2026.

We did skip the entire 2024 tax year, but fortunately, the bracket
numbers have no effect on the v1 API output. (Which raises the
question: is there any point in keeping this stuff once we delete v0?)

## Test Plan

Updated tests. I did a couple of the calculations manually to confirm;
trusting the rest are correct since I didn't touch the calculation
logic.

### DIFF
--- a/data/tax_brackets.json
+++ b/data/tax_brackets.json
@@ -2,197 +2,197 @@
   {
     "filing_status": "single",
     "income_min": 0,
-    "income_max": 11000,
+    "income_max": 11925,
     "tax_rate": 0.1,
-    "standard_deduction": 13850
+    "standard_deduction": 15000
   },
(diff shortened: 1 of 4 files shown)


oyamauchi:
For the record, there was a conversation in Slack about whether this is useful. The conclusion was that we probably want to filter out federal tax credits when we calculate zero tax liability, so we will want to keep this around and updated.

cjohnsbrown (approved):

PR #666, opened 2025-02-12
Add husky and lint-staged; rebuild sqlite DB after checkout

oyamauchi:
## Links

- [Asana](https://app.asana.com/0/1208923746898025/1206215453677154)

## Description

This brings this repo in line with our other ones, particularly
`consumer` and the embed repo. It does the same `prettier --write` on
commit that the other repos do, but notably, this one exclude the
`data` directory, because running Prettier on that takes a _long_ time
thanks to the tens of thousands of lines of JSON. Those files are not
hand-edited and are formatted as they're written, so we're not missing
out by skipping them here.

It's easy to forget to rebuild the SQLite database after its inputs
change, and that can result in strange-looking test failures. So add a
post-checkout hook that looks for changes to its inputs and rebuilds
it if there are any. (It takes about 1 second; it's not a major
burden.)

## Test Plan

On top of this state, make a change to a file in `src` and make sure
it gets prettified when I commit the change. Make a change to one of
the sqlite source CSV files and commit it; make sure the DB gets
rebuilt when checking out that commit and the one before it.

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,0 +1,12 @@
+set -e
+
+CHANGED=$(git diff "$1" "$2" --stat -- ./yarn.lock | wc -l)
+if (( CHANGED > 0 )); then
+    echo "📦 yarn.lock changed. Run yarn install to bring your dependencies up to date."
+fi
+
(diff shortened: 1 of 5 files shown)


oyamauchi:
I could claim I'm making a push to fix little tech debt type things, but no, I cannot tell a lie; the real reason I did this now is that I must have PR number 666.

cjohnsbrown (approved):

PR #667, opened 2025-02-13
Change order of filtering in state incentive calculation

oyamauchi:
## Description

This change does two things:

- Don't return ineligible incentives from
  `calculateStateIncentivesAndSavings`. Previously, they were being
  returned, and then filtered out in the top-level route handler in
  `v1.ts`. They were never returned in the API. This makes the
  below change simpler.

- Draw a distinction between "real-life eligibility" params, and
  "API-only" params. The latter are the `authority_types` and `items`
  filters. They don't correspond to real-life incentive eligibility
  criteria; they're just there for API clients.

  The important thing here is that now the order of operations in
  eligibility calculation is slightly different:

  1. All of a state's incentives are partitioned into "eligible" and
     "ineligible" based on real-life params: location, utility, owner
     status, income/filing/hhsize, and date. (Date is not an API
     param, but it's a fact of real life that has bearing on
     eligibility.) No incentives are completely removed from
     consideration at this point.

  2. Relationships are solved. This can rule out some incentives that
     came out of step 1 as "eligible".

  3. Incentives are removed from the eligible set based on API-only
     params.

  4. Total savings are calculated. (And almost immediately thrown out,
     and never returned through the API; I plan to remove this stuff
     separately.)

  Previously, filtering by API-only params happened at the top, before
  relationship solving; this gave those params unintended semantics.
  E.g. if you passed `items=ductless_heat_pump`, rather than meaning
  "just filter out incentives for items other than ductless heat
  pumps", it would _consider you ineligible_ for such incentives. That
  had undesired implications for relationship solving.

This is in preparation for implementing the Mass Save rule properly;
that will be a followup PR.

## Test Plan

`yarn test` with updated tests.

The changes to the snapshots are all expected. Those authorities were
in there even though none of the incentives belonged to them; it's
just that some _ineligible_ incentives belonged to them. Those
ineligible incentives' authorities were put into the response before
the incentives were filtered out.

### DIFF
--- a/test/snapshots/v0-80212-homeowner-80000-joint-4.json
+++ b/test/snapshots/v0-80212-homeowner-80000-joint-4.json
@@ -3,7 +3,7 @@
   "is_under_150_ami": true,
   "is_over_150_ami": false,
   "pos_savings": 14000,
-  "tax_savings": 5836,
+  "tax_savings": 5523,
   "performance_rebate_savings": 8000,
   "estimated_annual_savings": 1040,
   "pos_rebate_incentives": [
--- a/src/lib/state-incentives-calculation.ts
+++ b/src/lib/state-incentives-calculation.ts
@@ -17,6 +17,7 @@ import {
 import { AmountType } from '../data/types/amount';
 import { APICoverage } from '../data/types/coverage';
 import { OwnerStatus } from '../data/types/owner-status';
+import { Program } from '../data/types/program';
 import { APISavings, zeroSavings } from '../schemas/v1/savings';
 import { AMIAndEVCreditEligibility } from './ami-evcredit-calculation';
 import { lastDayOf } from './dates';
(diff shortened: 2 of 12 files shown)


test/snapshots/v0-80212-homeowner-80000-joint-4.json
```diff
@@ -3,7 +3,7 @@
   "is_under_150_ami": true,
   "is_over_150_ami": false,
   "pos_savings": 14000,
-  "tax_savings": 5836,
+  "tax_savings": 5523,
```
oyamauchi, line 6: This is actually a relic of #665. It turns out the v0 snapshot test wasn't actually testing the equality of the _entire_ response, so this snapshot test didn't fail on #665, so I didn't update the snapshots on that PR.

cjohnsbrown (approved):

PR #672, opened 2025-02-18
Fix import

oyamauchi:
## Description

The weekly HERO import didn't work, and it looks to be because of this
Husky post-checkout script. It turns out the newest version of Husky
runs hooks in `sh`, which does not support the double-paren-to-do-math
syntax that this hook was using (that's a Bash thing). Change it to be
sh-friendly.

## Test Plan

Will look at the CI logs on this PR to make sure there's no error
messages from this hook.

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,12 +1,12 @@
 set -e
 
 CHANGED=$(git diff "$1" "$2" --stat -- ./yarn.lock | wc -l)
-if (( CHANGED > 0 )); then
+if [ $CHANGED -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


oyamauchi:
OK well it turns out the CI log doesn't tell me the important piece of info. But I'm pretty sure this is the fix.

More detail: the proximate cause of the import failure was that there was a stray untracked file named `0` when the create-pull-request action ran. I'm pretty sure that got created by `sh` interpreting the `(( CHANGED > 0 ))` syntax as an i/o redirection to the file `0`.

PR #675, opened 2025-02-20
Calculate taxes for MA and NM

oyamauchi:
## Links

- [Asana](https://app.asana.com/0/1208668890181682/1209428558323182)

## Description

We have state tax credits for these, so we ought to calculate their taxes.

I learned through doing this that a lot of MA tax calculators on the
internet (or at least the ones that have good SEO) are wrong; they
either don't account for the millionaire's tax or use the wrong
threshold for it.

I also learned that Wikipedia has the best SEO for "new mexico state
tax" but the table on the page is wrong; it doesn't include the top
marginal rate of 5.9%, which is new in 2021.

I'm left with a general feeling of irritation that I'm not quite sure
what to do with.

## Test Plan

New unit tests. The NM expected values are the ones given in the
source document. The MA values were calculated by me actually going
through MA tax forms and following the instructions.

### DIFF
--- a/src/lib/tax-brackets.ts
+++ b/src/lib/tax-brackets.ts
@@ -25,44 +25,34 @@ const MA_SURTAX_THRESHOLD = 1_083_150;
 // https://www.tax.newmexico.gov/individuals/wp-content/uploads/sites/5/2021/12/PIT-rates_2005_2021.pdf
 const NM_BRACKETS: {
   [S in FilingStatus]: { maxIncome: number; rate: number }[];
-} = {
-  [FilingStatus.Single]: [
-    { maxIncome: 5500, rate: 0.017 },
-    { maxIncome: 11000, rate: 0.032 },
-    { maxIncome: 16000, rate: 0.047 },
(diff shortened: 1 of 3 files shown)


oyamauchi:
I updated the NM Wikipedia page, in case anyone goes to look and sees that it's correct now.

src/lib/tax-brackets.ts
```diff
@@ -4,11 +4,72 @@ import { TERRITORIES } from '../data/types/states';
...
+    { maxIncome: Infinity, rate: 0.059 },
+  ],
+  [FilingStatus.MarriedFilingSeparately]: [
+    { maxIncome: 4000, rate: 0.017 },
+    { maxIncome: 8000, rate: 0.032 },
+    { maxIncome: 12000, rate: 0.047 },
+    { maxIncome: 157500, rate: 0.049 },
+    { maxIncome: Infinity, rate: 0.059 },
+  ],
+  // Below three all the same
```
veekas, line 43: would it make sense to turn the arrays into a variable and use that three times?

veekas (approved):

PR #676, opened 2025-02-21
Filter out tax credits if we compute zero tax liability

oyamauchi:
## Links

- [Asana](https://app.asana.com/0/1208668890181682/1209381661213139)

## Description

Do this separately for federal and state incentives.

This also fixes a bug in the MA state tax calculation I just added,
which resulted in negative tax liability for incomes below the
personal exemption.

## Test Plan

New unit test.

### DIFF
--- a/src/lib/tax-brackets.ts
+++ b/src/lib/tax-brackets.ts
@@ -82,7 +82,7 @@ export function estimateStateTaxAmount(
     // gets the personal exemption automatically.)
     case 'MA': {
       const exemption = MA_EXEMPTIONS[filingStatus];
-      const taxableIncome = householdIncome - exemption;
+      const taxableIncome = _.max([0, householdIncome - exemption])!;
 
       // NB: the surtax is not a higher bracket of a graduated rate; it's
       // additional to the flat base rate.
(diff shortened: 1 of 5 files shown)


src/lib/tax-brackets.ts
```diff
@@ -82,7 +82,7 @@ export function estimateStateTaxAmount(
     // gets the personal exemption automatically.)
     case 'MA': {
       const exemption = MA_EXEMPTIONS[filingStatus];
-      const taxableIncome = householdIncome - exemption;
+      const taxableIncome = _.max([0, householdIncome - exemption])!;
```
veekas, line 85: 🐛 🔨

veekas (approved):

coopr:
> I don't love this - tax liability is reward looking (right? like "last year") while tax credits could apply if tax liability occurs this year (which hasn't finished yet, and until it has, you can't really know if there will be tax liability).
> 
> Better would be to include tax credits, with a warning that tax liability is needed to benefit from them.

Would love to hear back on this, thanks

RandomEtc:
> Would love to hear back on this, thanks

Can I ask for a little grace? Your question landed on a Friday afternoon for our east coast team, and while we are happy to have folks follow along with our work (it's open source because we want to share) perhaps a higher level discussion about API features and the scope of this product would be helpful? I'm not sure getting into this in a pull request will be productive. I'll email you!

PR #682, opened 2025-02-24
Add Incentives Endpoint for Incentives Visualizer

Adrastopoulos:
This PR adds adds new `/api/v1/incentives` endpoint that returns a list of incentives for use in the [incentives visualizer](https://github.com/rewiringamerica/incentives-visualizer).

- Supports optional state filtering via `?state=XX` query parameter
- Returns metadata including last updated timestamp and count of incentives/states

### DIFF
--- a/src/schemas/v1/incentives-endpoint.ts
+++ b/src/schemas/v1/incentives-endpoint.ts
@@ -1,0 +1,130 @@
+import { FromSchema } from 'json-schema-to-ts';
+import { AmountType } from '../../data/types/amount';
+import { PaymentMethod } from '../../data/types/incentive-types';
+import { ALL_ITEMS } from '../../data/types/items';
+import { OwnerStatus } from '../../data/types/owner-status';
+import { STATES_AND_TERRITORIES } from '../../data/types/states';
+
+export const API_INCENTIVES_REQUEST_SCHEMA = {
+  title: 'APIIncentivesRequest',
+  type: 'object',
+  properties: {
+    state: {
+      type: 'string',
+      description:
+        'Find incentives available in this state. If not provided, returns incentives from all states.',
+      enum: STATES_AND_TERRITORIES,
+    },
+  },
+  additionalProperties: false,
+} as const;
+
+export const API_INCENTIVES_RESPONSE_SCHEMA = {
+  $id: 'APIIncentivesResponse',
+  type: 'object',
+  required: ['incentives', 'metadata'],
+  properties: {
+    incentives: {
+      type: 'array',
+      items: {
+        type: 'object',
+        required: [
+          'id',
+          'items',
+          'short_description',
+          'payment_methods',
+          'amount',
+          'owner_status',
+          'program',
+        ],
+        properties: {
+          id: { type: 'string' },
+          items: {
+            type: 'array',
+            items: {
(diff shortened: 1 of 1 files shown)


Adrastopoulos:
We should consider replacing all `as const` keywords when declaring api schemas to `satisfies JsonSchema` for better type-safety.

src/schemas/v1/incentives-endpoint.ts
```diff
@@ -0,0 +1,130 @@
...
+            additionalProperties: false,
+          },
+          owner_status: {
+            type: 'array',
+            items: {
+              type: 'string',
+              enum: Object.values(OwnerStatus),
+            },
+          },
+          eligible_geo_group: { type: 'string' },
```
oyamauchi, line 89: In the JSON files, this field and `low_income` are IDs that reference whole structures stored in other JSON files. They aren't meaningful by themselves, so I don't think it makes sense to expose them in the API.

As background:

`eligible_geo_group` refers to one of the structures in `data/geo_groups.json`. A geo group is a set of authorities (which are cities, counties, or utilities). If an incentive has a geo group, you have to be "in" one of the geo group's authorities to be eligible (either your location is in one of the group's city- or county-type authorities, or you're a customer of one of the group's utilities). This constraint is in addition to the usual constraint that you must be "in" the authority that the incentive actually belongs to, unless that authority has `authority_type == 'other'`. (The function `ineligibleByLocationOrUtility` implements this logic.)

`low_income` refers to one of the structures in `data/low_income_thresholds.json`. Those structures describe requirements on household income to claim an incentive (usually that it must be <= $X for a household of size N).

So, for the purposes of this project:

- If you're just displaying incentives at state-level granularity, you don't need to worry about geo groups at all. They only affect eligibility at sub-state granularity. If you want to display incentives at finer granularity, it's going to get more complex and probably worth a separate discussion.

- It's probably may be enough to expose the low-income field as just a boolean: whether or not the `low_income` field is present in the raw incentive data. That would signify "is this incentive income-qualified or not?"

src/schemas/v1/incentives-endpoint.ts
```diff
@@ -0,0 +1,130 @@
...
+          low_income: { type: 'string' },
+        },
+        additionalProperties: false,
+      },
+    },
+    metadata: {
+      type: 'object',
+      required: ['last_updated', 'total_incentives', 'total_states'],
+      properties: {
+        last_updated: { type: 'string' },
```
oyamauchi, line 100: This information doesn't currently exist in the JSON data. We have talked about adding a last-updated timestamp to each incentive, but haven't gotten to it. You may want to drop it for now.

oyamauchi (commented):
Re: `as const` and `satisfies JSONSchema`. I think the benefit of adding the `satisfies` constraint isn't improved type safety, but better error messages. If the object wouldn't satisfy the `JSONSchema` type, you'll get a deeply-nested error in the `FromSchema` invocation. But by adding the constraint, you get a more-legible error at the exact point in the object that's wrong. And once you have the `satisfies` constraint, I believe `as const` doesn't make a difference either way.

That said, having better error messages is a real benefit. I don't mind adding `satisfies JSONSchema`.

PR #686, opened 2025-03-04
Add loan import script

benjf:
I tested this by temporarily adding a branch trigger in the `import.yml` workflow. After the last successful run, [this PR](https://github.com/rewiringamerica/api.rewiringamerica.org/pull/688) was created.

### DIFF
--- a/scripts/incentive-admin-capital-import.ts
+++ b/scripts/incentive-admin-capital-import.ts
@@ -22,9 +22,9 @@ async function saveLoanProgramDataToDisk() {
 
 async function fetchCapitalData(path: string) {
   const data_url = `${INCENTIVE_ADMIN_HOST}${path}`;
-  const authHeader = process.env.CAPITAL_API_KEY;
+  const authHeader = process.env.CAPITAL_ADMIN_API_KEY;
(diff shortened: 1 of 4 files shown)


scripts/incentive-admin-capital-import.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const GENERAL_IMPORTS = ['/loan-programs?status=PUBLISHED'];
+
+function pathToFileName(path: string) {
+  // Remove query params and slashes; replace dashes with underscores
+  return path.split('?')[0].replaceAll('/', '').replaceAll('-', '_') + '.json';
+}
+
+async function saveLoanProgramDataToDisk() {
+  for (const path of GENERAL_IMPORTS) {
```
rserur, line 15: wrap this in a try to catch any errors here?

rserur (approved):

scripts/incentive-admin-capital-import.ts
benjf: I figured the fetching would be most likely to fail, but probably good to also add a try/catch here around writing the file.

PR #700, opened 2025-03-13
Use geocodio type definitions

Adrastopoulos:
I've updated the geocodio node sdk to `1.7.0` which adds type definitions [here](https://github.com/Geocodio/geocodio-library-node/pull/39).

To pass type-checking, our code will either need runtime validation or non-null assertions (for example, with the census check)

### DIFF
--- a/src/lib/location.ts
+++ b/src/lib/location.ts
@@ -1,3 +1,4 @@
+import { GeocodeAccuracyType } from 'geocodio-library-node';
 import { Database } from 'sqlite';
 import { geocoder } from './geocoder';
 
@@ -15,7 +16,7 @@ export type ResolvedLocation = {
  * types are imprecise enough, in practice, that we can't trust the geocoded
  * census tract.
  */
-const GEOCODIO_LOW_PRECISION = new Set([
+const GEOCODIO_LOW_PRECISION = new Set<GeocodeAccuracyType>([
   'street_center',
   'place',
   'county',
@@ -47,13 +48,13 @@ export async function resolveLocation(
   } else {
     const { address } = zipOrAddress;
     const response = await geocoder.geocode(address, ['census2020']);
-    const result = response?.results?.at(0);
+    const result = response.results.at(0);
 
     if (!result || result.address_components.country !== 'US') {
       return null;
     }
 
-    const censusInfo = result.fields.census['2020'];
+    const censusInfo = result.fields?.census?.['2020'];
 
     return {
       state: result.address_components.state,
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "fastify-cli": "^6.1.1",
     "fastify-plugin": "^4.5.1",
     "fastify-sqlite": "^1.1.0",
-    "geocodio-library-node": "^1.6.1",
+    "geocodio-library-node": "^1.7.0",
     "lodash": "^4.17.21",
     "qs": "^6.11.2",
     "sqlite": "^5.1.1"
--- a/src/@types/geocodio-library-node.d.ts
+++ b/src/@types/geocodio-library-node.d.ts
@@ -1,2 +1,0 @@
-// Trivial module declaration to satisfy tsc
-declare module 'geocodio-library-node';
--- a/yarn.lock
+++ b/yarn.lock
@@ -2357,10 +2357,10 @@ generify@^4.0.0:
     split2 "^3.0.0"
     walker "^1.0.6"
 
-geocodio-library-node@^1.6.1:
-  version "1.6.1"
-  resolved "https://registry.yarnpkg.com/geocodio-library-node/-/geocodio-library-node-1.6.1.tgz#3762f6e3e9fc799652de67b310fb66596521398c"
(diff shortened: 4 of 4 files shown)


Adrastopoulos:
I made a [PR](https://github.com/Geocodio/geocodio-library-node/pull/43) on Geocodio's node sdk repo that fixes some mistakes in the type definitions. We should wait for that to be merged in before continuing here.

Adrastopoulos:
> I made a [PR](https://github.com/Geocodio/geocodio-library-node/pull/43) on Geocodio's node sdk repo that fixes some mistakes in the type definitions. We should wait for that to be merged in before continuing here.

I've sent an email to the owner to expedite this.

Adrastopoulos:
This is good to go @oyamauchi. 

From the owner of geocodio: "there's a few recent API changes that are not reflected in the OpenAPI spec yet."

I've added some `ts-expect-error` directives to handle this.

src/lib/location.ts
```diff
@@ -47,19 +49,24 @@ export async function resolveLocation(
...
 
     if (!result || result.address_components.country !== 'US') {
       return null;
     }
 
+    // @ts-expect-error census2020 is not included in the typedef
     const censusInfo = result.fields.census['2020'];
 
     return {
+      // @ts-expect-error state is optional in the typedef
```
rpartridge, line 62: rather than doing `@ts-expect-error` can you just assert that these values will be not null? ex `result.address_components.state!`

alternatively, if these types are saying these fields might be null, we probably just want to add error handling to handle the cases where these expected fields are null (even though in practice we know that's very unlikely to ever happen). defer to the incentives team on their preference here

src/lib/location.ts
```diff
@@ -47,19 +49,19 @@ export async function resolveLocation(
...
     const response = await geocoder.geocode(address, ['census2020']);
-    const result = response?.results?.at(0);
+    const result = response.results.at(0);
 
     if (!result || result.address_components.country !== 'US') {
       return null;
     }
 
+    // @ts-expect-error census.2020 is not included in the typedef
     const censusInfo = result.fields.census['2020'];
```
Adrastopoulos, line 59: I kept the expect-error directive here because the `Census` type does not have a `2020` property according to the type definitions.

src/lib/location.ts
Adrastopoulos: The non-null assertions are added.

rpartridge:
@oyamauchi @cjohnsbrown @veekas could one of y'all take a look at this when you get a chance?

src/lib/location.ts
veekas: Should it be? It [looks like it's a valid key](https://www.geocod.io/docs/#census-block-tract-fips-codes-amp-msa-csa-codes), and one of the contributors to the thread [here](https://github.com/Geocodio/geocodio-library-node/issues/17#issuecomment-2470818477) offered some type definitions for `Census`.

src/lib/location.ts
```diff
@@ -47,19 +49,19 @@ export async function resolveLocation(
...
 
     return {
-      state: result.address_components.state,
-      zcta:
-        (await zipLookup(db, result.address_components.zip))?.zcta ??
-        result.address_components.zip,
+      state: result.address_components.state!,
+      zcta: result.address_components.zip!,
+      // @ts-expect-error city is optional in the typedef
```
veekas, line 64: Can you share why you added non-null operators for `state` and `zip` but not `city`, and why you removed the `zipLookup` call?

veekas (commented):
Nice to have type definitions

src/lib/location.ts
Adrastopoulos: > Should it be? It [looks like it's a valid key](https://www.geocod.io/docs/#census-block-tract-fips-codes-amp-msa-csa-codes), and one of the contributors to the thread https://github.com/Geocodio/geocodio-library-node/issues/17#issuecomment-2470818477 offered some type definitions for Census.

You're absolutely correct. The issue is that these type definitions were written based off of the [openapi-sec](https://github.com/Geocodio/openapi-spec/blob/master/geocodio-api.yml), which erroneously maps `census` directly to `FieldCensus`, as opposed to a mapping of [`Fields`](https://github.com/Geocodio/openapi-spec/blob/master/geocodio-api.yml#L669) to [`FieldCensus`](https://github.com/Geocodio/openapi-spec/blob/master/geocodio-api.yml#L482).

We can address this (for now) by typecasting to the correct types here. Let me know what you think.

> Can you share why you added non-null operators for state and zip but not city, and why you removed the zipLookup call?

Oh. I think removing the zipLookup call was in error. I can add it back. The reason the `city` didn't have a non null operator was because of the ts-expect error directive. It is nullable just like `state` and `zip`, according to the openapi spec: https://github.com/Geocodio/openapi-spec/blob/master/geocodio-api.yml#L638

It is not immediately clear if these values are supposed to be nullable. Probably not, since we assumed that they were defined before adding these types. I've had some back and forth with the library author and he acknowledges that the openapi spec has quite a few inaccuracies, as it were. In any case, I'll add some typecasting and we can go from there.

veekas (approved):
thank you!

PR #703, opened 2025-03-20
Update language to match the zuplo documentation

benjf:


src/schemas/v1/loan-programs.ts
```diff
@@ -154,7 +154,8 @@ export const API_LOAN_PROGRAMS_REQUEST_SCHEMA = {
  */
 export const API_LOAN_PROGRAMS_ENDPOINT_SCHEMA = {
   summary: 'Get all loan programs by location',
-  description: 'Returns all loan programs for a specified location.',
+  description:
+    "Returns the relevant loan programs for a location. The location can be specified by either a ZIP code or a full address (one of these parameters is required, providing both we result in an error). The filtering is currently just based on a user's State.",
```
SkittyWitty, line 158: Maybe adding more details about what the loan programs are for? Idea if someone is new to the API and is trying to understand how to use it. 

Another option is running this script [here](https://github.com/rewiringamerica/api-zuplo/blob/main/scripts/merge-spec.ts) on the `spec.js` and generating return responses and to add to the [API Responses](https://api.rewiringamerica.org/docs/routes/responses).

SkittyWitty (approved):

PR #718, opened 2025-04-08
Add solar water heating IRA incentive

veekas:
## Description

Adds an IRA tax credit for solar water heaters, which is available under 25D.

Note: I _didn't_ update the v0 endpoint because it seemed like that wasn't a requirement in the [previous ticket](https://github.com/rewiringamerica/api.rewiringamerica.org/pull/699) but if reviewers feel like I should add it, I'll do so.


veekas:
I didn't mean to click Copilot as a requested reviewer, but it was interested to see what it can/not review.

PR #722, opened 2025-04-17
Adding in 60% AMI Limits

tarrenpeterson:
See notion [here](https://www.notion.so/rewiringamerica/AMI-MFI-for-Incentives-a18db19ff12840319472bc2803a63b31?pvs=4#1d54c9af1a1c80038732f02088ed6d24) for details on the decision to implement a 60% AMI level for our GGRF programs.

This PR updates the income_level scripts and data to add a 60% AMI column by multiplying the 50% AMI column by 1.2.


scripts/income_limits/data/processed/ami_by_territory_zip.csv
mikivee, line 1: It looks like the numbers changed...? Is this pulling 2025 data? I know this gets updated sometime in April 🤔

tarrenpeterson: Oooh good call - I did not think of that. Looks like 2025 data was published on April 1st. Do you think we should rerun with 2024 data?

mikivee: I think this is an @oyamauchi question. The plan was to do a data refresh as described in the readme every April.... so maybe you just did Owen's work for him? but should def let him decide if now is a good time to change all the numbers

oyamauchi:
When to switch to 2025 numbers is currently an open question for us, actually: https://app.asana.com/1/1200412452784804/project/1208668890181682/task/1209928851911935

Basically the problem is that some programs say you must be under 80% AMI or whatever, but we don't necessarily know when _they_ switch to the new year's HUD numbers. That is, when someone in the program administrator's office is actually reviewing a rebate application, which AMI table are they looking at? Do they all switch on April 1 precisely? Probably not.

I don't feel equipped to answer this question myself, on this PR. How hard would it be to rerun with 2024 numbers, so we can sidestep it for now and unblock you?

Also, I'm probably missing something, but how do you plan to use these numbers downstream? As this PR is, they won't affect anything in the API eligibility calculation logic.

tarrenpeterson:
> When to switch to 2025 numbers is currently an open question for us, actually: https://app.asana.com/1/1200412452784804/project/1208668890181682/task/1209928851911935
> 
> Basically the problem is that some programs say you must be under 80% AMI or whatever, but we don't necessarily know when _they_ switch to the new year's HUD numbers. That is, when someone in the program administrator's office is actually reviewing a rebate application, which AMI table are they looking at? Do they all switch on April 1 precisely? Probably not.
> 
> I don't feel equipped to answer this question myself, on this PR. How hard would it be to rerun with 2024 numbers, so we can sidestep it for now and unblock you?
> 
> Also, I'm probably missing something, but how do you plan to use these numbers downstream? As this PR is, they won't affect anything in the API eligibility calculation logic.

It's quite easy to rerun with 2024 numbers - I anticipated we might need to do so and reran locally during the all hands. I just added the 2024 files but it's still showing a large number of line-level changes. I'll take a closer look at the CSVs to see what's going on. 

We use the [raw CSVs](https://raw.githubusercontent.com/rewiringamerica/api.rewiringamerica.org/refs/heads/main/scripts/income_limits/data/processed/ami_by_tract.csv) in a couple different places (like [this PR](https://github.com/rewiringamerica/building_profile/pull/21)) to assess market size for GGRF programs.

mikivee (approved):
Lgtm! Thanks for adding this in. Looks like the numbers are unchanged now. 

Good luck with the 🧶 tests 😬

oyamauchi:
The test failures are because the new column in the CSVs isn't accounted for in the process that loads the CSVs into a SQLite database for the server to use at runtime.

You'll need to edit `scripts/import-csvs.sql`, to update some CREATE TABLE statements with the new column. Then, if you're running the JS tests locally, you'll need to run `yarn build` first (which actually builds the SQLite database). If you're not running the tests locally and just letting CI do it, you shouldn't need to do anything else.

tarrenpeterson:
> The test failures are because the new column in the CSVs isn't accounted for in the process that loads the CSVs into a SQLite database for the server to use at runtime.
> 
> You'll need to edit `scripts/import-csvs.sql`, to update some CREATE TABLE statements with the new column. Then, if you're running the JS tests locally, you'll need to run `yarn build` first (which actually builds the SQLite database). If you're not running the tests locally and just letting CI do it, you shouldn't need to do anything else.

Thank you for the suggestion - I was at a loss for how to debug those yarn tests!

PR #724, opened 2025-04-18
Import geographies from HERO; generate mapping from ZCTAs to geos

oyamauchi:
## Links

- [Asana](https://app.asana.com/1/1200412452784804/project/1208668890181682/task/1209816248106961)

## Description

Depends on rewiringamerica/incentive_admin#259.

- Import the geographies table from HERO as CSV, and add the optional
  `geography_id` field to authorities, to reference them.

  I decided to import the geographies as CSV, rather than JSON,
  because all the geometry data makes the file awkwardly large. As
  CSV, it can go into the SQLite database and thus not be loaded into
  memory all at once on startup.

- Add a script that generates a mapping from ZCTAs to geography IDs
  that they intersect, and load that into the runtime SQLite database.

  My original plan was for the API server to have geometry for ZCTAs
  available at runtime, and compute intersections with geographies at
  query time.

  However, "geometry for ZCTAs" turns out to mean hundreds of
  megabytes, which is a little excessive for this context, so this is
  the alternative: precompute those intersections. This results in a
  CSV file of less than 5 megabytes, which is roughly the size it ends
  up as in SQLite; this is totally tolerable.

  We only need to recompute this when geographies change in HERO. I
  don't anticipate that happening very often, so I'm OK with running
  this script manually in the near term. I've added a unit test that
  will fail if there are any geographies with no intersections with
  ZCTAs (other than two for which this is expected), which is an
  indication that this script needs to be rerun.  Automatically
  running this script during HERO imports will require figuring out
  how to get the census data files automatically; I'd rather handle
  that in a separate task/PR.

## Test Plan

Spot-checked a few zips I'm familiar with, as well as a few
cross-state ones:

- 38079 (contains the Kentucky Bend, spanning KY and TN)
- 97635 (spans CA and OR)

Spot-checked some of the other county-spanning zips on the map, to see
whether they really do span counties, and whether the proportion in
the CSV seems right. They do seem to be real -- I'm finding addresses
with the ZIP on both sides of the state or county border.


scripts/incentive-admin-import.ts
```diff
@@ -46,11 +51,12 @@ async function saveIncentiveDataToDisk() {
...
   }
 
-  // zip-to-utility.csv
-  const writePath = 'scripts/data/zip-to-utility.csv';
-  const data = await fetchZipToUtility();
-  writeFileSync(writePath, data);
-  console.log(`Exported ${writePath}...`);
+  for (const path of CSV_IMPORTS) {
+    const data = await fetchCsv(path);
+    const writePath = `scripts/data/${path.replaceAll('/', '')}.csv`;
```
cjohnsbrown, line 56: nit: If you take out the `/` after `data` you can remove the replaceAll

cjohnsbrown (approved):

PR #726, opened 2025-04-23
Find matching geographies at query time

oyamauchi:
## Description

1. If the input is a zip, look up geography IDs in the
   zcta-to-geography table. If the input is an address, send it to
   Geocodio, use the returned state and county to pick geographies for
   those, and use the returned lat/long to compute intersection with
   every custom geography. (This sounds like a lot of work but the
   time it takes is dominated by the Geocodio call.)

2. For incentive eligibility, compare the resolved geography IDs
   against the geography ID of the incentive's authority, or against
   the geography IDs of the incentive's geo group's member
   authorities.

3. For program eligibility, same as above but without the geo group
   part.

3. For county-dependent income thresholds, take the minimum value for
   each household size across all resolved counties.

One important note is that we're currently returning any incentive
whose authority's geography intersects _at all_ with any of the
geographies the user _might_ be in. This is pretty permissive, but is
in line with how things worked before this change. We'll certainly
want to change this going forward, probably in conjunction with some
way for the API to indicate conditional eligibility.

- This changes the computed eligibility of some city and county
  incentives.  It may seem disruptive, but that's the whole point of
  this project. In any event, it's much more sound than using postal
  cities as we were doing before.

- The `city` and `county_fips` fields of the location object in the
  API response are now never populated. (This is what all the snapshot
  changes are about.) The fields are officially optional, so this does
  not break the API contract. The calculator also doesn't use them for
  anything.

  I'd rather not try to populate these fields; they're fundamentally
  based on an assumption that the user's location resolves to a single
  city/county, which isn't the case anymore. Instead I want to (a)
  consider what location fields we want to expose in the API at all,
  and (b) design a useful way to expose them.

## Test Plan

New and updated unit tests.

Test some state-spanning ZCTAs: 97635 (spans CA and OR; should resolve
to CA), 02861 (spans MA and RI; should resolve to RI); make sure they
resolve to the right state.

Test some ZCTAs that should get city or county incentives, and make
sure those city/county incentives are returned: 80310 (City of Boulder
and Boulder County), 80510 (Boulder County only), 06351 (part of CT
"distressed municipalities"; should get a $450 battery incentive "for
homeowners in underserved communities), 06018 (CT, should not get that
battery incentive).

Some fun little test cases for matching from a geocoded address:

- `2077 S State St, Ann Arbor, MI 48104`. This should NOT return any
  City of Ann Arbor incentives (the A2Zero program). Change the street
  number to 2051 (right next door) and it should return A2Zero
  incentives. 2077 is in one of the many "township islands" of Ann
  Arbor: plots of land that were never absorbed into the city.

- `1808 Church St, Evanston, IL 60201`. This should NOT return any
  Evanston Development Cooperative incentives. Change the street
  number to 1815 (right across the street) and then it should. This
  street is a boundary of the Evanston Green Homes pilot area, a
  custom geography.


cjohnsbrown (approved):
Do we have documentation for the sources of the edge-case custom geographies?

oyamauchi:
I came up with those edge cases myself, by looking around at the boundaries of custom geos, and just by looking for city/county incentives.

PR #732, opened 2025-04-28
add error handling for geocode error

veekas:
## Description

- [Asana](https://app.asana.com/1/1200412452784804/project/1208668890181682/task/1209940781479960?focus=true)
- [Github conversation](https://github.com/rewiringamerica/embed.rewiringamerica.org/pull/231#discussion_r2038071911)

Screenshot 2025-04-28 at 1 45 11 PM

### DIFF
--- a/src/lib/location.ts
+++ b/src/lib/location.ts
@@ -1,5 +1,6 @@
 import * as turf from '@turf/turf';
 import { Database } from 'sqlite';
+import { InvalidInputError } from './error';
 import { geocoder } from './geocoder';
 
 export enum GeographyType {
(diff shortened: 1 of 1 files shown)


src/lib/location.ts
```diff
@@ -107,80 +108,87 @@ export async function resolveLocation(
...
-
-    // There is a relatively small number of 'custom' geographies, so running
-    // the "contains" computation over all of them is quick. If necessary, we
-    // can speed this up by using a spatial index and/or caching the parsed
-    // GeoJSON in memory.
-    (
-      await db.all<DbGeography[]>(
-        `SELECT * FROM geographies WHERE type = 'custom'`,
+    try {
+      const response = await geocoder.geocode(address, ['census2024']);
```
oyamauchi, line 112: This is the only line that needs to be inside the try block; this line is what's throwing the exception. With everything else in the try block, any other exceptions that get thrown will get treated as a generic failure to resolve the location, which may end up hiding errors we would rather know about.

src/lib/location.ts
```diff
@@ -107,80 +108,87 @@ export async function resolveLocation(
...
+          (await getZctaFromZip(db, result.address_components.zip)) ??
+          result.address_components.zip,
+        geographies,
+        tract_geoid: GEOCODIO_LOW_PRECISION.has(result.accuracy_type)
+          ? undefined
+          : censusInfo.county_fips + censusInfo.tract_code,
+      };
+    } catch (error) {
+      throw new InvalidInputError(
+        `Unable to geocode address: ${address}`,
```
oyamauchi, line 189: So I was kinda wrong when I explained the plan before. The problem is that `resolveLocation` is called from the top level code in `v1.ts` outside of any try blocks. Its (implicit) contract is that it never throws; it indicates error by returning null. (This is why the visible text in the screenshot still says "internal server error"; it's still serving a 500.)

If you just return null here, the right thing should happen: v1.ts will return a localized error message, with a 404 status, that says we can't locate that address.

oyamauchi (changes requested):

src/lib/location.ts
```diff
@@ -107,7 +107,14 @@ export async function resolveLocation(
     };
   } else {
     const { address } = zipOrAddress;
-    const response = await geocoder.geocode(address, ['census2024']);
+    let response;
+
+    try {
+      response = await geocoder.geocode(address, ['census2024']);
+    } catch (error) {
+      return null;
```
oyamauchi, line 115: Let's log the error here with `console.error`; it currently doesn't go anywhere except Cloud Run logs but it could be useful to have.

oyamauchi, line 115: Let's log the error here with `console.error`; it currently doesn't go anywhere except Cloud Run logs but it could be useful to have.

oyamauchi (approved):

PR #733, opened 2025-04-28
Validate OpenAPI spec in test; fix errors

oyamauchi:
## Links

- [Asana](https://app.asana.com/1/1200412452784804/project/1208668890181682/task/1209935388457419)

## Description

There were some things about the endpoint schemas that were not valid
OpenAPI.

- `type` has to be a string. A value that is "string or null" has to
  have `type: string, nullable: true`.

- `if` and `then` aren't allowed. (I moved this logic so that it still
  runs against the JSON file, but isn't part of the API spec.)

- A default value of an incorrect type.

In general, OpenAPI spec is **not** a superset of JSON Schema! Even
though it looks like it. Specifically, look here:

https://spec.openapis.org/oas/v3.0.3.html#properties

> Additional properties defined by the JSON Schema specification that are not mentioned here are strictly unsupported.

This fixes those issues, and adds a unit test that validates the
produced spec. It was surprisingly hard to find a library that would
do what I wanted here; this one is not exactly ideal (suboptimal docs,
no TS types) but it's the best I found; the most important thing is
that it produces usable error messages.

### Followup

- Merge the updated spec into the Zuplo config, so the changes get into the public spec.

- Maybe implement a similar validation test in the Zuplo repo.

## Test Plan

Run tests. To make sure the if/then conditional validations are done,
I changed a `is_national: false` to true in `loan_programs.json`, and
made sure the test suite fails.


benjf (approved):
Thanks Owen!

PR #734, opened 2025-04-29
Add support for "% of federal poverty level" LI thresholds

oyamauchi:
## Links

- [Asana](https://app.asana.com/1/1200412452784804/project/1208668890181682/task/1209254256140597)

## Description

This is a pretty simple scheme: there are hhsize-indexed tables that
go up to 8, plus a defined amount for each additional person. There
are tables for Alaska and Hawaii, and another for the lower 48 + DC.
There's no FPL defined in the territories.

### Next steps

Add support for these in HERO; export to API.

## Test Plan

Unit test.


cjohnsbrown (approved):

veekas (approved):
only nit is that idk if "other" makes sense to me as the category that applies to all lower 48 states, but that's not actually important lol

PR #755, opened 2025-06-03
Use Github App for HERO import

oyamauchi:
## Links

- [Jira](https://rewiringamerica.atlassian.net/browse/RAT-568)

## Description

This has two upsides over the previous approach of simply using the
provided `GITHUB_TOKEN`:

- CI will run without requiring a manual step
- It can add the incentives team as a reviewer

The app is all set up already, including configuring the two secrets
on this repo. The secrets are actually set at the organization level
so the same app can easily be used in other repos for similar
processes (i.e. anything that requires authenticating to Github but
where `GITHUB_TOKEN` isn't sufficient for whatever reason).

## Test Plan

Ran locally using https://github.com/nektos/act . It's not a perfect
simulation of Github itself, but it did work; the result is
rewiringamerica/api.rewiringamerica.org#754 .

### DIFF
--- a/.github/workflows/import.yml
+++ b/.github/workflows/import.yml
@@ -11,6 +11,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Generate token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.IMPORTER_APP_ID }}
+          private-key: ${{ secrets.IMPORTER_PRIVATE_KEY }}
       - name: Checkout repository
         uses: actions/checkout@v4
       - name: Use Node.js 18.x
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -11,6 +11,7 @@ build
 scripts/income_limits/data/raw/
 scripts/income_limits/config.json
 secrets/
+.secrets
```
oyamauchi, line 14: This file is how you pass secrets to Act (workflow testing tool)

.github/workflows/import.yml
```diff
@@ -31,8 +37,6 @@ jobs:
       - name: Check for changes
         id: git_diff
         run: |
-          git config --global user.name 'github-actions[bot]'
-          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
```
oyamauchi, line 35: I think this had no effect anyway. The create-pull-request action sets the author and committer on its own, and without these the PR creator is shown as `hero-importer (bot)`, which is what I wanted.

cjohnsbrown (approved):

PR #756, opened 2025-06-04
Remove periods in schema $id

oyamauchi:
## Links

- [Jira](https://rewiringamerica.atlassian.net/browse/RAT-366)

## Description

The new merge script that @SkittyWitty wrote (rewiringamerica/api-zuplo#140) trips up on schema names that have periods in them (because it uses lodash to follow period-delimited paths through a JSON object). Rather than try to deal with that in the script, I just fixed the schema names; there's no reason to have the `.schema.json` part.

## Test Plan

`yarn test`. Fetch /spec.json locally and diff it with the spec generated from main to make sure removing the string `.schema.json` is the only difference.


SkittyWitty:
Thanks for being proactive @oyamauchi ! ✨

cjohnsbrown (approved):

PR #785, opened 2025-07-15
Update end dates of IRA tax credits

oyamauchi:
## Links

- [Jira](https://rewiringamerica.atlassian.net/browse/RAT-686)

## Description

- 25C (energy efficient home improvement) and 25D (residential clean
  energy) are available for installations that happen before the end
  of 2025.

- 30D (new EV) and 25E (used EV) are available for vehicles purchased
  before the end of September.

- 30C is available for installations before the end of June 2026.

Note that in the v0 API, the end date is just a number signifying the
year, so these are returned in v0 as just `2025` (or `2026` in one
case). v0 is about to be deleted, so I'm not worried about that.

As written, this means that we'll completely stop showing them after
the end date. However, people who want to claim them will still need
to remember to do so, come tax time next year; I'm not sure if we'll
want to keep showing these cards in 2026 for that reason?

## Test Plan

`yarn test`

### DIFF
--- a/data/ira_incentives.json
+++ b/data/ira_incentives.json
@@ -21,7 +21,7 @@
     ],
     "ami_qualification": null,
     "start_date": "2023",
-    "end_date": "2032",
+    "end_date": "2025-12-31",
     "short_description": {
       "en": "30% tax credit (uncapped) for battery storage systems, worth $4,800 on average.",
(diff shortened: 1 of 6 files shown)


cjohnsbrown (approved):
😕

PR #791, opened 2025-07-24
Treat IRA incentives like state incentives

oyamauchi:
## Links

- [Jira](https://rewiringamerica.atlassian.net/browse/RAT-711)

## Description

Ever since state-level incentives were introduced, they've been
treated differently from IRA incentives: a different data type, fed
into an entirely different function to filter them for eligibility.

This fixes that situation: IRA incentives are now of the
`StateIncentive` type, and there's a single function that evaluates
eligibility for all incentives. This allows for a lot of code
simplification.

Another advantage is that we can represent the income qualifications
of the EV credits using the same "low income thresholds" logic as
state incentives, instead of the awkward situation of having copies of
a single incentive that differ only in the `agi_max_limit` field or
whatever. I added the 30D and 25E brackets in HERO a while back, and
use them here.

Since only the IRA tax credits were ever returned in v1, and not the
rebates, I deleted all the rebate from the JSON.

(I know all the IRA incentives are ending soon, and I could have just
waited and deleted all the "federal" code at once, but look... I
wanted a little treat _now_, and deleting thousands of lines of code
felt like just the thing.)

## Test Plan

`yarn test`.


test/lib/incentives-calculation.test.ts
```diff
@@ -216,9 +214,9 @@ test('correctly evaluates scenario "Single w/ $120k Household income in NJ"', as
     i => i.payment_methods[0] === PaymentMethod.PerformanceRebate,
   );
 
-  t.equal(pos_rebate_incentives.length, 7);
-  t.equal(tax_credit_incentives.length, 12);
-  t.equal(performance_rebate_incentives.length, 1);
+  t.equal(pos_rebate_incentives.length, 0);
+  t.equal(tax_credit_incentives.length, 10);
```
oyamauchi, line 218: The number went down because the "state" calculation function doesn't return ineligible incentives at all, rather than returning them with `eligible: false`. The end result in the v1 API is the same.

test/lib/incentives-calculation.test.ts
```diff
@@ -289,15 +258,12 @@ test('correctly evaluates scenario "Single w/ $120k Household income in NJ"', as
   t.equal(taxCredits['heat_pump_water_heater'].start_date, '2023');
   t.equal(taxCredits['rooftop_solar_installation'].eligible, true);
   t.equal(taxCredits['rooftop_solar_installation'].amount.number, 0.3);
-  t.equal(taxCredits['rooftop_solar_installation'].amount.representative, 4356);
```
oyamauchi, line 292: This was based on some really old solar cost estimates and the number was never shown in the current calculator anyway.

oyamauchi (commented):
Couple of comments to preemptively address questions

src/lib/state-incentives-calculation.ts
```diff
@@ -248,10 +277,13 @@ function ineligibleByLocationOrUtility(
     return true;
   }
 
-  const authority =
-    stateAuthorities[program.authority_type][program.authority!];
+  // Federal incentives have no authority, and need no geography check
+
+  const authority = program.authority
```
cjohnsbrown, line 282: Would it be more clear to use `program.authority_type === AuthorityType.Federal` as the condtion?

cjohnsbrown (approved):
Thanks for giving some additional context in the comments

PR #796, opened 2025-07-29
Fix bug in coverage reporting

oyamauchi:
## Description

This was caused by the big post-v0 overhaul I just did.
`calculateStateIncentive` was computing the `coverage` part of the API
response by whether the list of incentives passed to it was empty. In
the overhaul, I made it so that it's _never_ empty (it always includes
IRA incentives), so the returned `coverage` was always non-null.

This changes it to compute `coverage` one level up, which allows for
some code simplification.

## Test Plan

New snapshot test that failed before the change. Existing tests pass.


matelau (approved):
👍🏼

PR #897, opened 2026-04-10
feat: release all states

susanafotu:
# Overview

This PR moves all states in BETA to LAUNCHED, effectively releasing all incentive data we have out into the API. 

## Change Type
Feature

## Implementation Notes
To minimize scope creep and to not cause breaking changes to the API, I chose to not remove include_beta_states from API schema or remove StateStatus.Beta from the enum. 

## Test Plan

Validate you can pull incentives for CA.


2WheelCoder (approved):
LGTM and CA incentives came down!

PR #918, opened 2026-07-23
bump fastify

matelau:
# Overview
Upgrades Fastify and its companion plugins from major v4 to v5. The previous Dependabot bump of `fastify` alone (#901) failed CI because `@fastify/type-provider-json-schema-to-ts` was still pinned to a v2 release built for Fastify 4's type-provider interface, which broke `request.query` type inference across `src/routes/v1.ts`. This bumps the whole ecosystem together so the types line back up.

## Issue Ticket
[Ticket](https://rewiringamerica.atlassian.net/browse/PDM-277)

## Change Type
Dependency upgrade (chore)

## Implementation Notes
- `fastify` ^4.17.0 → ^5.8.5
- `@fastify/type-provider-json-schema-to-ts` ^2.2.2 → ^4.0.1
- `@fastify/autoload` ^5.8.2 → ^6.4.0
- `@fastify/cors` ^9.0.1 → ^10.1.0
- `@fastify/sensible` ^5.6.0 → ^6.0.4
- `@fastify/swagger` ^8.15.0 → ^9.8.1
- `fastify-cli` ^6.3.0 → ^7.4.1
- `fastify-plugin` ^4.5.1 → ^5.1.0
- `fastify-tsconfig` ^2.0.0 → ^3.0.0

No application code changed — the existing `.withTypeProvider<JsonSchemaToTsProvider<...>>()` call in `src/routes/v1.ts` is already valid against the new type-provider version once `fastify` and the provider are on matching majors.

## Test Plan
1. `yarn lint` — tsc, prettier, eslint all pass with no errors.
2. `yarn build && yarn test` — full suite passes (30,681 assertions, 1 skip, 0 failures).
3. `yarn dev`, then manually hit each endpoint to confirm runtime behavior beyond what the test suite covers:

   ```bash
   # /api/v1/states
   curl -s http://localhost:3000/api/v1/states

   # /api/v1/calculator — happy path
   curl -s "http://localhost:3000/api/v1/calculator?zip=80212&owner_status=homeowner&household_income=80000&tax_filing=joint&household_size=4"

   # /spec.json — OpenAPI spec still generates
   curl -s http://localhost:3000/spec.json

   # /api/v1/calculator — 400 shape (household_size=0 is invalid)
   curl -s "http://localhost:3000/api/v1/calculator?zip=80212&owner_status=homeowner&household_income=80000&tax_filing=joint&household_size=0"

   # /api/v1/calculator — 404 shape (zip doesn't exist)
   curl -s "http://localhost:3000/api/v1/calculator?zip=00000&owner_status=homeowner&household_income=80000&tax_filing=joint&household_size=4"
   ```

   The 400/404 responses came back as `{"statusCode":400,"error":"Bad Request","message":"querystring/household_size must be >= 1"}` and `{"statusCode":404,"error":"Not Found","message":"That ZIP code doesn't exist."}` respectively — unchanged from pre-upgrade behavior.


2WheelCoder:
One thing worth noting from review: `@fastify/type-provider-json-schema-to-ts@^4.0.1` pulls in `json-schema-to-ts@^3.1.0` as a transitive dependency, but the project's own devDependency is still pinned at `json-schema-to-ts@^2.0.0` (resolves to 2.9.1). That direct v2 package is what's actually imported for `FromSchema` throughout the codebase (`src/schemas/v1/*.ts`, `src/data/*.ts`, `src/data/types/*.ts` — 16 files), while the type-provider used in `src/routes/v1.ts` now does its internal type inference against v3.

Both resolve fine today and `tsc`/tests are clean, so this isn't blocking. But it leaves two majors of the same inference engine in the tree, which is a latent foot-gun: a future schema change could get differently-inferred types depending on whether it flows through the direct v2 `FromSchema` or the type-provider's internal v3 one, and TS wouldn't necessarily flag the mismatch loudly. Worth bumping the direct `json-schema-to-ts` devDependency to `^3.1.0`+ in a fast follow to collapse back to one version.

matelau:
@2WheelCoder 
Addressed in the latest commit: bumped the direct `json-schema-to-ts` devDependency `^2.0.0` → `^3.1.1`, so the tree now resolves to a single `json-schema-to-ts@3.1.1` (shared with the type-provider's transitive copy).

v3 makes defaulted schema properties required in inferred types, which broke direct lib calls in `test/lib/low-income.test.ts` and `test/lib/programs-for-location.test.ts`. To keep the dedup behavior-neutral, passed `{ keepDefaultedPropertiesOptional: true }` to the three request-type `FromSchema` aliases (`APICalculatorRequest`, `APIProgramsRequest`, `APILoanProgramsRequest`), preserving the prior v2 typing. Response types have no `default:` fields, so they were untouched.

`yarn lint` and `yarn test` both clean (30,681 pass / 1 skip / 0 fail).

2WheelCoder:
> @2WheelCoder Addressed in the latest commit: bumped the direct `json-schema-to-ts` devDependency `^2.0.0` → `^3.1.1`, so the tree now resolves to a single `json-schema-to-ts@3.1.1` (shared with the type-provider's transitive copy).
> 
> v3 makes defaulted schema properties required in inferred types, which broke direct lib calls in `test/lib/low-income.test.ts` and `test/lib/programs-for-location.test.ts`. To keep the dedup behavior-neutral, passed `{ keepDefaultedPropertiesOptional: true }` to the three request-type `FromSchema` aliases (`APICalculatorRequest`, `APIProgramsRequest`, `APILoanProgramsRequest`), preserving the prior v2 typing. Response types have no `default:` fields, so they were untouched.
> 
> `yarn lint` and `yarn test` both clean (30,681 pass / 1 skip / 0 fail).

Ah, nice. I [dove this work while you were out](https://github.com/rewiringamerica/api.rewiringamerica.org/pull/919), but forgot to pull it all the way through into this PR. Thanks for bringing it home and sorry we had to duplicate some work.

2WheelCoder (approved):

PR #919, opened 2026-07-24
Bump json-schema-to-ts to ^3.1.0

2WheelCoder:
# Overview
Follow-up from review of #918: `@fastify/type-provider-json-schema-to-ts@^4.0.1` (added in #918) already pulls in `json-schema-to-ts@3.1.1` transitively, but the project's direct devDependency was still pinned at `^2.0.0` — leaving two majors of the same type-inference library in the tree. This bumps the direct dependency to `^3.1.0` so both resolve to the same version.

json-schema-to-ts v3 infers schema properties with a `default` (but not listed in the schema's `required` array) as **required** rather than optional, since AJV will always populate them before a handler runs. This tightened `CalculateParams`/`APIProgramsRequest` around `language`/`include_beta_states`, breaking ~40 test fixtures across 4 test files that previously omitted those fields at the type level.

No production/runtime behavior change — this only affects TypeScript compile-time inference in tests. Real API behavior is governed by the JSON schema `default` + AJV's `useDefaults`, both unchanged by this PR.

## Issue Ticket
[Ticket](https://rewiringamerica.atlassian.net/browse/PDM-277)

## Change Type
Dependency upgrade (chore)

## Implementation Notes
- `json-schema-to-ts` ^2.0.0 → ^3.1.0 (devDependency)
- Added shared `DEFAULT_CALCULATE_PARAMS` (`test/mocks/calculate-params.ts`) and a local `DEFAULT_PROGRAMS_REQUEST` constant, both defined via `satisfies Partial<T>` rather than `as const` — this validates the fixture against the real type without forcing every property to `readonly`.
- Test fixtures across `incentives-calculation.test.ts`, `incentive-relationships.test.ts`, `low-income.test.ts`, and `programs-for-location.test.ts` now spread these shared defaults instead of omitting the defaulted fields, matching the values AJV would apply at runtime. This also deduplicated several repeated literal blocks (5x in `incentive-relationships.test.ts`, 7x in `programs-for-location.test.ts`).

## Test Plan
1. `yarn lint` — tsc, prettier, eslint all pass with no errors.
2. `yarn build && yarn test` — full suite passes (30,681 assertions, 1 skip, 0 failures) — matches #918's baseline exactly.


matelau:
@2WheelCoder I think we can close this one now right, since it was related to the other PR that was merged in.

